PR #9, opened 2016-02-17
Adding/updating migrations for the database structure.

### LINKED ISSUE #2: db structure
Create the following migrations:

- [x] users
- [x] ~~roles~~
- [x] ~~role_user~~
- [x] waiting_room [id | nid | run_nid | signup_dates]
- [x] competitions  [id | nid | run_nid | comp_dates]
- [x] competition_user

weerd:
Fixes #2

#### What's this PR do?
This PR updates/add migrations to setup the database structure that we discussed as a team.

Check out the migrations in the **/database** directory.

Pull down the branch and run `php artisan migrate` and make sure you don't get any errors! If you already had the old migrations setup, make sure to `php artisan migrate:reset` or clear out your database before pulling down the branch!

#2 

#### Questions:
Are there any additional fields that should be treated as `->index()` for lookups, etc? (specifically looking at you @angaither :wink: )

Also, apart from the `northstar_id` in the `users` table, should we also have one for `phoenix_id`?

---
@angaither @sbsmith86 

cc: @DFurnes @deadlybutter


database/migrations/2016_02_17_175810_create_competition_user_table.php
```diff
@@ -0,0 +1,33 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('competition_user', function (Blueprint $table) {
+            $table->increments('id');
+            $table->integer('competition_id')->unsigned()->index();
+            $table->foreign('competition_id')->references('id')->on('competitions')->onDelete('cascade');
```
sbsmith86: @weerd does this mean that if you delete a competition user you would also delete the competition from the competitions table?

DFurnes:
Noting an IRL conversation – we should have a pivot table for the `WaitingRoom ↔ User` many-to-many relationship too!

database/migrations/2016_02_09_224824_create_competitions_table.php
```diff
@@ -14,9 +14,10 @@ public function up()
     {
         Schema::create('competitions', function (Blueprint $table) {
             $table->increments('id');
-            $table->date('start_date');
-            $table->date('end_date');
+            $table->integer('campaign_id')->unsigned()->index();
             $table->integer('campaign_run_id')->unsigned()->index();
+            $table->date('start_at');
```
angaither: i kind of like still calling these `start_date` `end_date`

database/migrations/2014_10_12_000000_create_users_table.php
```diff
@@ -14,10 +14,8 @@ public function up()
     {
         Schema::create('users', function (Blueprint $table) {
             $table->increments('id');
```
angaither: do we need an id on this? can the northstar id be the real id? and have an index?

database/migrations/2016_02_17_175810_create_competition_user_table.php
```diff
@@ -0,0 +1,33 @@
...
+class CreateCompetitionUserTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('competition_user', function (Blueprint $table) {
```
angaither: hm

database/migrations/2016_02_17_175810_create_competition_user_table.php
weerd: @sbsmith86 I think it's more if you delete a competition user it will delete the reference in the pivot table for the competitions that user was associated with.

database/migrations/2016_02_09_224824_create_competitions_table.php
weerd: yeah, I was debating that a bit, cause then in the waiting_room table would it be `signup_start_date` and `signup_end_date`? Also was trying to mimic the naming that the timestamps create... `created_at` & `updated_at`... I can change though if no one likes it :)

database/migrations/2014_10_12_000000_create_users_table.php
weerd: Not entirely sure if maybe the app would benefit from it's own user id? I noticed that @DFurnes has the same structure in [Aurora](https://github.com/DoSomething/aurora/blob/8a3c5319b0b67cf7e8ec8789ce875ea4ab0661e9/database/migrations/2015_01_15_213047_create_users_table.php#L16) and was trying to follow a similar approach, so maybe he can chime in to see if it ended up benefitting to have separate id for app along w/ NS_id...?

database/migrations/2016_02_17_175810_create_competition_user_table.php
weerd: pivot table! we need this and the `user_waiting_room` pivot table which I'm adding as well :)

weerd:
@DFurnes updated and added the `user_waiting_room` migration :)

database/migrations/2014_10_12_000000_create_users_table.php
angaither: yeah, when i made aurora it was just a part of the user table automatically, and I think I needed it b/c i had the roles pivot table since it worked as the main id. I don't think it's needed or used though

weerd: okie doke. I'll remove it. Should I add a `phoenix_id` at all? Would that help when signups from Phoenix for competitions are hitting Gladiator and bringing with it data from Phoenix?

weerd: @angaither Ok, so seems like we need the app id that's an integer for the pivot table stuff that relies on ids to do the associations, otherwise the migrations and artisan yell at me a lot...

database/migrations/2016_02_09_224824_create_competitions_table.php
angaither: I'm pretty sure the standard is for date fields to explicitly say the word date in them
where `at` is more about when content is created/edited/removed

deadlybutter: I agree with calling it start_date and end_date -- thats how we also named it in the hot module

database/migrations/2016_02_17_175810_create_competition_user_table.php
```diff
@@ -0,0 +1,33 @@
...
+            $table->increments('id');
+            $table->integer('competition_id')->unsigned()->index();
+            $table->foreign('competition_id')->references('id')->on('competitions');
```
weerd: @sbsmith86 chatted further with @DFurnes and think you're correct on being concerned it might delete a user when competition is deleted. I was kind of following the examples with some old code on Aurora, but turns out that isn't even used anymore. Removed the `->onDelete('cascade')` so I can look into it further, and might be that those methods need to be on the individual migration tables and not on the pivots. Will keep you posted once I learn more ;)

database/migrations/2016_02_09_224824_create_competitions_table.php
weerd: sounds good!

angaither:
kewl kewl! :ship:

weerd:
Working on the Models and relationships next so we can seed the DB!

PR #10, opened 2016-02-18
Removing incremental ids from pivots and fixing bug.

weerd:
#### What's this PR do?
This PR does a quick update to remove the incremental id column on the pivot tables since they're unnecessary, and fixes a typo on `down()` method for the `waiting_rooms` table.

---
@angaither @sbsmith86 

cc: @DFurnes


angaither:
:+1:

Issue #17, opened 2016-02-25
Add validation errors on login


weerd:
Login authentication via Northstar needs to return list o' errors and show them to the user when logging in.

PR #20, opened 2016-02-26
Adds validation messages appropriate to the Northstar response.

weerd:
#### What's this PR do?
This PR updates the validation so it can send the custom error messages that Northstar responds with to properly indicate what the issue was with the authentication. This can also be customized further depending on additional error responses that Northstar can return with.

Start at the **RestApiClient.php** file and follow the method trail from there!

Try logging in with bogus infoz and you should see error messages.

When email is correct but password is wrong, should say `Invalid credentials.`.

#17 

---

### DIFF
--- a/app/Services/RestApiClient.php
+++ b/app/Services/RestApiClient.php
@@ -4,6 +4,8 @@ namespace Gladiator\Services;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\RequestException;
+use Illuminate\Support\MessageBag;
+use Illuminate\Contracts\Validation\ValidationException;
 
 class RestApiClient
 {
@@ -56,7 +58,7 @@ class RestApiClient
             'query' => $query,
         ]);
 
-        return is_null($response) ? null : $this->getJson($response);
+        return is_null($response) ? null : $this->getJson($response)->data;
     }
 
     /**
@@ -72,7 +74,7 @@ class RestApiClient
             'body' => $this->makeJson($body),
         ]);
 
-        return is_null($response) ? false : $this->getJson($response);
+        return is_null($response) ? false : $this->getJson($response)->data;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


weerd:
Need to try a quick fix so I don't `use` the `ValidationException` contract but the actual class instead... brb

app/Services/RestApiClient.php
```diff
@@ -97,7 +99,7 @@ public function put($path, $body = [])
      */
     public function getJson($response, $asArray = false)
     {
-        return json_decode($response->getBody(), $asArray)->data;
+        return json_decode($response->getBody(), $asArray);
```
angaither: interesting, why remove it from here?

weerd: For the errors... if we get an error response `data` doesn't exist... instead you have to request `->error` or even `->errors` ;) Better to let where it's actually being used grab what it needs and let the `getJson()` just turn the response into Json but not decide what property might contain the info you want.

resources/views/auth/login.blade.php
```diff
@@ -18,12 +18,12 @@
 
                     <div class="form-item -padded">
                         <label class="field-label" for="email">Email</label>
-                        <input type="text" name="email" id="email" class="text-field" placeholder="kallark&#64;dosomething.org" required>
+                        <input type="text" name="email" id="email" class="text-field" placeholder="kallark&#64;dosomething.org">
```
angaither: haha to avoid © infringement?

angaither:
:+1:

resources/views/auth/login.blade.php
weerd: lol. nope, so I can show nice customized validation errors with our own Forge interface instead of the default HTML5 popup :)

angaither: oh i meant the name kallark 
thought you were going for clark kent
for reasons i now see aren't that rational...

PR #24, opened 2016-02-29
Fixing small but important bug in WaitingRoom model.

weerd:
#### What's this PR do?
This PR fixes a bug/typo that was affecting the proper response when using eager loading in Laravel's Eloquent. Need to make sure to return object so that it can be used when chaining methods.

Take a wild guess ;)

This was affect eager loading in Eloquent so we couldn't do things like:
```php
WaitingRoom::with('users')->findOrFail('3');
```

---
@angaither @deadlybutter @sbsmith86

### DIFF
--- a/app/Models/WaitingRoom.php
+++ b/app/Models/WaitingRoom.php
@@ -13,6 +13,6 @@ class WaitingRoom extends Model
      */
     public function users()
     {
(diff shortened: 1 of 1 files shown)


deadlybutter:
:shipit:

PR #34, opened 2016-02-29
Commiting work to show diego

deadlybutter:


deadlybutter:
@weerd if you wanna take a look :eyes:

app/Http/routes.php
```diff
@@ -27,6 +27,8 @@
     Route::resource('competitions', 'CompetitionsController');
 
     // Waiting rooms routes.
+    Route::get('waitingrooms/{waitingrooms}/split', 'WaitingRoomsController@showSplitForm')->name('split');
+    Route::post('waitingrooms/{waitingrooms}/split', 'WaitingRoomsController@split')->name('split');
```
weerd: I've never seen it like you have it... only like this:

```
Route::get('waitingrooms/{waitingrooms}/split', ['as' => 'waitingrooms.split', 'uses' => 'WaitingRoomsController@showSplitForm']);
```

deadlybutter: hmm ill try that then

deadlybutter: nope @weerd still getting this error `Missing required parameters`

deadlybutter: fixed the route to look like that for both the post & get, also updated the views

Issue #32, opened 2016-02-29
User CRUD


weerd:
Falls in same boat with methods related to #22

Issue #22, opened 2016-02-26
Create endpoint for Phoenix to post users


weerd:
When a user confirms that they would like to sign up for a competition, Phoenix sends a post request, and need to implement an endpoint on Gladiator to receive this post and add the new user to the `User` tables and place in appropriate waiting room.

PR #35, opened 2016-03-01
Adding User finding helper functions

weerd:
#### What's this PR do?
This PR adds some helper functions to help deal with handling finding a user within the DS system, whether they're in Northstar, in Northstar but not Gladiator, or in both Northstar and in Gladiator. The aim is to create some functions that can be called that handle a lot of the User lookup throughout the system and help avoid repeated code throughout the app.

here's an example of how you could use the helper function:
```php
$user = findUserAccountInSystem('dlorenzo@dosomething.org', 'email');

if ($user instanceof Gladiator\Models\User) {
  // We found the user in system as Gladiator user, so lets just add them to a WaitingRoom...
}

// User was found on Northstar (since no exception thrown), but not on Gladiator.
// All we get is a NS id string, lets go ahead and add them as a new user to Gladiator!
$northstarId = $user->id;

$user = new User;
$user->id = $northstarId;
$user->save();
```

There are also functions for `findNorthstarAccount()` and `findGladiatorAccount()` which are referenced in the `findUserInSystem()` function which is more of a shortcut that does both all in one shot.

Um, best way to test for now is to set up a test route an try calling the methods with appropriate info.

I went a bit back and forth on this one. Originally I had these functions as methods of the User model, but when actually using them it all felt weird to have to instantiate the User class to call these functions to even decide if a user already existed. Ideally methods on the User model would be used to deal with a loaded user object from the database and not used to call a bunch of helper methods. I ended up deciding to go with the helper global functions approach. Alternatively, I was also considering maybe a Utilities class that could be called and loaded and these could be methods on that class, but helper functions felt a little easier.

Refs #22 
Refs #32 

--- 
@DFurnes @sbsmith86 @deadlybutter

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -1,0 +1,74 @@
+<?php
+
+use Gladiator\Models\User;
+use Gladiator\Services\Northstar\Exceptions\NorthstarUserNotFoundException;
+
+if (! function_exists('matchEmailDomain')) {
+    /**
+     * Match supplied email against a specific domain or default domain.
+     *
+     * @param  string  $email
+     * @param  string  $domain
+     * @return boolean
+     */
+    function matchEmailDomain($email, $domain = 'dosomething.org')
+    {
+        $pieces = explode('@', $email);
+
+        if (is_array($pieces) && count($pieces) > 1) {
+            return $pieces[1] === $domain ? true : false;
+        }
+
+        return false;
+    }
+}
+
+if (! function_exists('findUserInSystem')) {
+    /**
+     * @param  string  $id
+     * @param  string  $email
+     * @return \Gladiator\Models\User|string
+     * @throws \Gladiator\Services\Northstar\Exceptions\NorthstarUserNotFoundException
+     */
+    function findUserInSystem($id, $type = 'email')
+    {
+        $northstarUser = findNorthstarAccount($id, $type);
+
+        if (! $northstarUser) {
+            throw new NorthstarUserNotFoundException;
+        }
+
+        $gladiatorUser = findGladiatorAccount($northstarUser->id);
+
+        if (! $gladiatorUser) {
+            return $northstarUser->id;
+        }
+
+        return $gladiatorUser;
+    }
+}
+
+if (! function_exists('findGladiatorAccount')) {
+    /**
+     * @param  string  $id
+     * @return \Gladiator\Models\User|null
+     */
+    function findGladiatorAccount($id)
+    {
+        return User::find($id);
+    }
+}
+
+if (! function_exists('findNorthstarAccount')) {
+    /**
+     * @param  string  $id
+     * @param  string  $type
+     * @return object|null
+     */
+    function findNorthstarAccount($id, $type = 'email')
+    {
+        $northstar = app('northstar');
+
+        return $northstar->getUser($id, $type);
+    }
+}
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -3,72 +3,64 @@
 use Gladiator\Models\User;
 use Gladiator\Services\Northstar\Exceptions\NorthstarUserNotFoundException;
 
-if (! function_exists('matchEmailDomain')) {
-    /**
-     * Match supplied email against a specific domain or default domain.
-     *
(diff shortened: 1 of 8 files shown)


app/helpers.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+use Gladiator\Models\User;
+use Gladiator\Services\Northstar\Exceptions\NorthstarUserNotFoundException;
+
+if (! function_exists('matchEmailDomain')) {
```
DFurnes: You probably don't need this in a "app" helpers directory, since you wouldn't ever want to accidentally use an existing global function here.

weerd: oh gotcha. was following Laravel's approach, but yeah, good point.

app/helpers.php
```diff
@@ -0,0 +1,74 @@
...
+}
+
+if (! function_exists('findUserInSystem')) {
+    /**
+     * @param  string  $id
+     * @param  string  $email
+     * @return \Gladiator\Models\User|string
+     * @throws \Gladiator\Services\Northstar\Exceptions\NorthstarUserNotFoundException
+     */
+    function findUserInSystem($id, $type = 'email')
```
DFurnes: Hmm, having these as global functions feels off to me. Would this make sense as a static method on the `User` model?

app/helpers.php
```diff
@@ -0,0 +1,74 @@
...
+        return $gladiatorUser;
+    }
+}
+
+if (! function_exists('findGladiatorAccount')) {
+    /**
+     * @param  string  $id
+     * @return \Gladiator\Models\User|null
+     */
+    function findGladiatorAccount($id)
```
DFurnes: Especially here... seems to sacrifice readability.

app/helpers.php
weerd: I was thinking about that as well, wasn't sure what was best and also for maintaining testability...

DFurnes:
My inline comments got squashed from the new commit, but I'm not sure these make sense as global functions. Have you tried adding this as a static method on the `User` class?

weerd:
yeah, wasn't sure why your comments weren't showing up. Anyhoo, I was thinking about them as static methods as well, wasn't sure what was best and also for maintaining testability...

app/helpers.php
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * @param  string  $id
+ * @param  string  $type
+ * @return object|null
+ */
+function findNorthstarAccount($id, $type = 'email')
+{
+    $northstar = app('northstar');
+
+    return $northstar->getUser($id, $type);
```
DFurnes: Is this backwards? `$northstar->getUser($type, $id)`?

DFurnes: Oh never mind, I see it's switched in this app's `Northstar` class.

weerd: Nope. I changed it up in my [Northstar](https://github.com/DoSomething/gladiator/blob/eaed9352a1909323c2ca3564add3149edeeb9309/app/Services/Northstar/Northstar.php#L30) client ;)

weerd: I wanted to default to a specific type so I don't always have to declare it.

sbsmith86:
This might be a dumb question, but what should this stuff live outside of the Northstar service class?

app/helpers.php
DFurnes: Okay, although I think we should try to keep at least similar behavior for methods (even if we're not using the shared Composer package here). :grimacing:

weerd: I'll be switching it up soon in favor of the NS package.

DFurnes: Cool cool. We could consider adding a `getUserByEmail` shortcut if you find you're using this method all the time... the two parameters are just to keep things in sync with the real endpoints.

weerd:
@sbsmith86 mostly cause it involves more than just NS and while it's using NS to check if user has NS account, it needs to deal with if they already exist in Gladiator. Also to help unify that combined logic to dry it up so it's not repeated in a few places.

weerd:
@DFurnes I'm trying it out as static methods on the User model and can push up a different PR.

PR #36, opened 2016-03-01
User static methods

weerd:
#### What's this PR do?
This PR adds some static helper method to help deal with handling finding a user within the DS system, whether they're in Northstar, in Northstar but not Gladiator, or in both Northstar and in Gladiator. The aim is to create some static methods that can be called that handle a lot of the User lookup throughout the system and help avoid repeated code throughout the app.

here's an example of how you could use the helper function:
```php
$user = Gladiator\Models\User::hasAccountInSystem('email', 'dlorenzo@dosomething.org');

if ($user instanceof Gladiator\Models\User) {
  // We found the user in system as Gladiator user, so lets just add them to a WaitingRoom...
}

// User was found on Northstar (since no exception thrown), but not on Gladiator.
// All we get is a NS id string, lets save the id and add them as a new user to Gladiator!
$northstarId = $user->id;

$user = new User;
$user->id = $northstarId;
$user->save();
```

Um, best way to test for now is to set up a test route an try calling the static methods with appropriate info.

In #35 I added these as global helper functions since I didn't like them as methods on the User model, but I wasn't super convinced that was a good idea. After a bit of discussion in that PR, tried going the route of static methods as @DFurnes suggested and I definitely like it better!

Also @DFurnes I fixed up the order of the params on the `getUser()` NS method so it matches when I switch over to use the composer package :)

Refs #22 
Refs #32 

--- 
@DFurnes @sbsmith86 @deadlybutter

### DIFF
--- a/app/Console/Commands/AddUser.php
+++ b/app/Console/Commands/AddUser.php
@@ -35,13 +35,13 @@ class AddUser extends Command
             return $this->comment(PHP_EOL . $email . ' is invalid. Admin or Staff require a DoSomething.org email.' . PHP_EOL);
         }
 
-        $northstarUser = findNorthstarAccount($email, 'email');
+        $northstarUser = User::hasNorthstarAccount('email', $email);
 
         if (is_null($northstarUser)) {
             return $this->comment(PHP_EOL . 'No user found on Northstar with email: ' . $email . '. Create an account at https://dosomething.org.' . PHP_EOL);
         }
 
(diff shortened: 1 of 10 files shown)


sbsmith86:
This looks good to me :)

weerd:
Cool! thanks @sbsmith86

DFurnes:
Post-merge :+1:!

PR #37, opened 2016-03-01
Adds ability to split the waiting room

### LINKED ISSUE #6: Split the waiting room
After the signup dates in a waiting room has ended an admin needs to be able to split users out into competitions

deadlybutter:
#### What's this PR do?
Lets the admin split a waiting room into multiple competitions

If the waiting room signup end date has passed, does it show a split button on the waiting room index and show view?

screen shot 2016-03-01 at 1 39 32 pm

Does it correctly and evenly split users into multiple groups on the preview? (You might need to change the number [here](https://github.com/DoSomething/gladiator/compare/master...split#diff-640369e4b888b6a682eafa65f228cf99R32) since our seed does not make 300+ users for a waiting room) 

screen shot 2016-03-01 at 1 47 33 pm

Are you able to successfully split the competition into multiple competitions based on the competition form details? 

screen shot 2016-03-01 at 1 48 22 pm

Does it update the results in the pivot tables?
screen shot 2016-03-01 at 1 49 04 pm
screen shot 2016-03-01 at 1 50 10 pm
screen shot 2016-03-01 at 1 50 43 pm
screen shot 2016-03-01 at 1 51 12 pm
screen shot 2016-03-01 at 1 51 26 pm

Fixes #6

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -2,7 +2,7 @@
 
 namespace Gladiator\Http\Controllers;
 
-use Illuminate\Http\Request;
+use Gladiator\Http\Requests\CompetitionRequest;
 use Gladiator\Models\Competition;
 
 class CompetitionsController extends Controller
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -7,7 +7,7 @@ use Gladiator\Models\Competition;
 
 class CompetitionsController extends Controller
 {
-    private $validationRules = [
+    public static $validationRules = [
         'campaign_id' => 'required|numeric',
         'campaign_run_id' => 'required|numeric',
         'start_date' => 'required|date',
@@ -113,5 +113,12 @@ class CompetitionsController extends Controller
         $competition->delete();
 
         return redirect()->route('competitions.index');
+    }
+
+    /**
+     * Returns the competition form validators.
+     */
+    public static function getValidationRules() {
+        return self::$validationRules;
     }
 }
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -7,13 +7,6 @@ use Gladiator\Models\Competition;
 
 class CompetitionsController extends Controller
 {
-    public static $validationRules = [
-        'campaign_id' => 'required|numeric',
-        'campaign_run_id' => 'required|numeric',
-        'start_date' => 'required|date',
-        'end_date' => 'required|date',
-    ];
-
     /**
      * Create new CompetitionsController instance.
      */
@@ -51,12 +44,9 @@ class CompetitionsController extends Controller
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function store(Request $request)
+    public function store(StoreCompetitionRequest $request)
     {
-        $this->validate($request, $this->validationRules);
-
         Competition::create($request->all());
-
         return redirect()->route('competitions.index');
     }
 
@@ -93,7 +83,7 @@ class CompetitionsController extends Controller
      * @param  \Gladiator\Models\Competition  $competition
      * @return \Illuminate\Http\Response
      */
-    public function update(Request $request, Competition $competition)
+    public function update(StoreCompetitionRequest $request, Competition $competition)
     {
         $this->validate($request, $this->$validationRules);
 
@@ -113,13 +103,5 @@ class CompetitionsController extends Controller
         $competition->delete();
 
         return redirect()->route('competitions.index');
-    }
-
-    /**
-     * Returns the competition form validators.
-     */
-    public static function getValidationRules()
-    {
-        return self::$validationRules;
     }
 }
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -118,7 +118,8 @@ class CompetitionsController extends Controller
     /**
      * Returns the competition form validators.
      */
-    public static function getValidationRules() {
+    public static function getValidationRules()
+    {
         return self::$validationRules;
     }
 }
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -44,7 +44,7 @@ class CompetitionsController extends Controller
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function store(StoreCompetitionRequest $request)
+    public function store(CompetitionRequest $request)
     {
         Competition::create($request->all());
         return redirect()->route('competitions.index');
@@ -83,7 +83,7 @@ class CompetitionsController extends Controller
      * @param  \Gladiator\Models\Competition  $competition
      * @return \Illuminate\Http\Response
      */
-    public function update(StoreCompetitionRequest $request, Competition $competition)
+    public function update(CompetitionRequest $request, Competition $competition)
     {
         $this->validate($request, $this->$validationRules);
 
--- a/app/Http/Controllers/WaitingRoomsController.php
+++ b/app/Http/Controllers/WaitingRoomsController.php
@@ -124,7 +124,7 @@ class WaitingRoomsController extends Controller
     public function showSplitForm(WaitingRoom $room)
     {
         $split = $room->getDefaultSplit();
-        
+
         return view('waitingrooms.split', compact('split', 'room'));
     }
 
@@ -135,10 +135,8 @@ class WaitingRoomsController extends Controller
      * @param  \Gladiator\Models\WaitingRoom  $room
      * @return \Illuminate\Http\Response
      */
-    public function split(Request $request, WaitingRoom $room)
+    public function split(StoreCompetitionRequest $request, WaitingRoom $room)
     {
-        $this->validate($request, CompetitionsController::getValidationRules());
-
         $split = $room->getDefaultSplit();
         $room->saveSplit($request->all(), $split);
 
--- a/app/Http/Controllers/WaitingRoomsController.php
+++ b/app/Http/Controllers/WaitingRoomsController.php
@@ -119,7 +119,8 @@ class WaitingRoomsController extends Controller
         return view('waitingrooms.split', compact('split', 'room'));
     }
 
-    public function split(WaitingRoom $room, Request $request) {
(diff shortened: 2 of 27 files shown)


deadlybutter:
gonna fix the styleci stuff now

app/Http/Controllers/CompetitionsController.php
```diff
@@ -7,7 +7,7 @@
 
 class CompetitionsController extends Controller
 {
-    private $validationRules = [
+    public static $validationRules = [
```
weerd: this feels weird...

app/Models/WaitingRoom.php
```diff
@@ -15,4 +16,62 @@ public function users()
...
         return $this->belongsToMany(User::class);
     }
+
+    /*
+     * Equally splits the users of a waiting room into arrays
+     * representing competitions.
+     */
+    public function getDefaultSplit()
+    {
+        $users = DB::table('user_waiting_room')->where('waiting_room_id', $this->attributes['id'])->get();
```
deadlybutter: ah i forgot to update this with eloquent

app/Http/Controllers/CompetitionsController.php
deadlybutter: I thought variables had to be static to be referenced by a static function?

app/Http/routes.php
```diff
@@ -27,6 +27,8 @@
     Route::resource('competitions', 'CompetitionsController');
 
     // Waiting rooms routes.
+    Route::get('waitingrooms/{waitingrooms}/split', 'WaitingRoomsController@showSplitForm')->name('split');
+    Route::post('waitingrooms/{waitingrooms}/split', 'WaitingRoomsController@split')->name('split');
```
weerd: did it not work using `['as' => 'split', 'use' => 'WaitingRoomsController@split']` etc? I feel this is more readable as to naming the route, rather than calling a name function. But if it doesn't bug anyone else, then not a huge deal...

app/Models/WaitingRoom.php
weerd: Did the eager loading I mentioned in Slack not work?

resources/views/waitingrooms/index.blade.php
```diff
@@ -36,6 +37,11 @@
                             <td class="table__cell">{{ $room->campaign_run_id }}</td>
                             <td class="table__cell">{{ date('F d, Y', strtotime($room->signup_start_date)) }}</td>
                             <td class="table__cell">{{ date('F d, Y', strtotime($room->signup_end_date)) }}</td>
+                            @if (time() - (60 * 60 * 24) >= strtotime($room->signup_end_date))
```
weerd: maybe add this as a global helper function you can call in the newly created **app/helpers.php** file from my last PR?

app/Http/routes.php
deadlybutter: It might have? I changed that (to what you suggested) while I was debugging something else. I didn't feel like it was a problem since it did the same thing, but ¯\_(ツ)_/¯

app/Models/WaitingRoom.php
deadlybutter: no it does I just forgot to change it

app/Http/Controllers/CompetitionsController.php
weerd: yeah, it's more about the needing to reference validation for Competitions in the WaitingRoomController.... that's the part that feels weird.

resources/views/waitingrooms/index.blade.php
weerd: Pull down latest master, rebase with this branch and run `composer dump-autoload` and you should be able to access the helper functions.

app/Models/WaitingRoom.php
weerd: gotcha!

app/Http/Controllers/CompetitionsController.php
deadlybutter: Not really sure of any better way to do this 

And from an admin perspective (Chris, Hillary) having the competition form on the split page is the most convenient way to do this.

weerd: It's really only the date assignment on the competition that matters for reasoning to have the form on the same page right?

Just feels very bad-practice-y to have the same form accessible from two different areas... maybe it should just default dates and then the admin can click on the competition to edit it directly and that'll take them to the `/competitions/:id/edit` page would be expected?

deadlybutter: I'm obviously new to this so I'm not sure why its bad practice, especially since the form is a partial. If we're making forms partials then I'm assuming it's built to be used in more than one place? 

If we did what you proposed there, Chris/Hillary could be going back and editing multiple competitions potentially if they didn't like the default dates. I feel like saving them a few minutes of time might be worth the small compromise we're making here.

A possible alternative? Can we make a function for creating/saving competitions, that both the competition controller & waiting room controller use?

DFurnes: @deadlybutter @weerd I think this would be a good scenario for a shared [Form Request](https://laravel.com/docs/5.2/validation#form-request-validation) that you could use from both controllers.

deadlybutter: mmm this looks like a good idea

weerd: Yeah, I had mentioned wanting to move to using FormRequests for validation stuff before... now might be a good time as @DFurnes suggests ;)

app/Http/Controllers/WaitingRoomsController.php
```diff
@@ -4,6 +4,7 @@
 
 use Illuminate\Http\Request;
 use Gladiator\Models\WaitingRoom;
+use Gladiator\app\Http\Requests\CompetitionRequest;
```
DFurnes: This should be `use Gladiator\Http\Requests\CompetitionRequest;`

deadlybutter: Oh that makes sense. I think I should probably get an IDE for Laravel dev

app/Http/Controllers/CompetitionsController.php
```diff
@@ -51,12 +44,9 @@ public function create()
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function store(Request $request)
+    public function store(CompetitionRequest $request)
```
DFurnes: You need to import this class into this file with a `use` statement. Otherwise it'll be looking for it in the current namespace (e.g. `Gladiator\Http\Controllers\CompetitionsRequest`).

`use Gladiator\Http\Requests\CompetitionRequest;`

app/Http/Controllers/WaitingRoomsController.php
weerd: and add it to the top of the CompetitionsController as well.

deadlybutter:
@weerd ready for another review!

app/Models/WaitingRoom.php
```diff
@@ -15,4 +15,62 @@ public function users()
...
+     */
+    public function getDefaultSplit()
+    {
+        $users = $this->users;
+
+        // Get the size of the waiting room
+        $roomSize = count($users);
+
+        // Determine the amount of competitions to make
+        $numOfCompetitions = $roomSize / 300;
```
weerd: So what happens if `$roomSize` is `305`? You get a decimal number of competitions; should you `floor` or `ceil` it? Also, do we get a room with 5-ish people?

deadlybutter: If the roomsize is a decimal nothing breaks,

and no we would never get a room with 5 people. Its always an even split (so 305.5 would have a 152 and 153 group or whatever the math is)

app/Models/WaitingRoom.php
```diff
@@ -15,4 +15,62 @@ public function users()
...
+     * representing competitions.
+     */
+    public function getDefaultSplit()
+    {
+        $users = $this->users;
+
+        // Get the size of the waiting room
+        $roomSize = count($users);
+
+        // Determine the amount of competitions to make
```
weerd: Not important for commit, but don't think we need that many comments on all these steps ;) Just leave comments on the ones that might seem a little cryptic. :wink:

resources/views/competitions/partials/_form_competitions.blade.php
```diff
@@ -2,12 +2,12 @@
...
 <div class="form-item -padded">
     {!! Form::label('campaign_id', 'Campaign ID:', ['class' => 'field-label']) !!}
-    {!! Form::text('campaign_id', NULL, ['class' => 'text-field']) !!}
+    {!! Form::text('campaign_id', (isset($room->campaign_id)) ? $room->campaign_id : NULL, ['class' => 'text-field']) !!}
 </div>
 
 <div class="form-item -padded">
     {!! Form::label('campaign_run_id', 'Run ID:', ['class' => 'field-label']) !!}
-    {!! Form::text('campaign_run_id', NULL, ['class' => 'text-field']) !!}
+    {!! Form::text('campaign_run_id', (isset($room->campaign_run_id)) ? $room->campaign_run_id : NULL, ['class' => 'text-field']) !!}
```
weerd: Was there a specific reason you had to add this back in? This should be autopopulating via the From Model binding with the correct info.

app/Models/WaitingRoom.php
weerd: Cool cool. Just double checking.

deadlybutter: :+1:  i had this fear early on when doing the math on paper but in practice it didnt prove to be a problem, so perfectly valid concern!

app/Models/WaitingRoom.php
deadlybutter: yeah, i feel like people often say I forget comments so i went overboard on this one haha

resources/views/competitions/partials/_form_competitions.blade.php
deadlybutter: not sure how it could auto bind if it was null (and in practice it wasnt, remember when you walked over I pointed out I had to do this?)

deadlybutter: Also the form wasnt being opened on an existing competition, so not sure how to make it bind anyway.

weerd: So the Form Model binding will leave it as `null` if it doesn't have any data to use otherwise it injects the data. Maybe check the docs or see how we're binding via the routes?

weerd:
:+1: just the note on the model-binding but not crucial and can be revisited later.

PR #38, opened 2016-03-01
User CRUD via interface & API requests

### LINKED ISSUE #32: User CRUD

weerd:
# Work In Progress

#### What's this PR do?
This PR adds User CRUD interface and setups up start for user persistence in the database via POST request.

User Index
image

User Create
image

User Show
image

User Edit
image

Pull down the code, run `composer dump-autoload` and log in. Try viewing the users list, editing their role or creating new ones.

Based on an IRL conversation with @DFurnes I moved the API request storing out of the `UsersController` to place in more dedicated API related controllers since I expected to run into some trouble with middleware and how web middleware is applied to the `UsersController` and how that would affect API requests. So the API user storing is coming in a separate PR.

Fixes #32 
Refs #22 

---

cc: @DFurnes

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -10,15 +10,26 @@ use Gladiator\Http\Requests\UserRequest;
 class UsersController extends Controller
 {
     /**
+     * Create new UsersController instance.
+     */
+    public function __construct()
+    {
+        $this->middleware('auth');
+        $this->middleware('role:admin,staff');
+    }
+
+    /**
      * Display a listing of the resource.
      *
      * @return \Illuminate\Http\Response
      */
     public function index()
     {
-        $users = User::all();
+        $admins = User::where('role', '=', 'admin')->get();
+        $staff = User::where('role', '=', 'staff')->get();
+        $contestants = User::where('role', '=', null)->get();
 
-        return view('users.index', compact('users'));
+        return view('users.index', compact('admins', 'staff', 'contestants'));
     }
 
     /**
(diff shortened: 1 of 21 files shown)


resources/views/users/partials/_form_users.blade.php
```diff
@@ -0,0 +1,41 @@
+@include('layouts.errors')
+
+{{ csrf_field() }}
+
+<div class="form-item -padded">
+    <label class="field-label" for="key">User identification key:</label>
+    <input type="text" name="key" id="key" class="text-field" placeholder="Email, Mobile Phone Number, Northstar ID or Phoenix ID" value="{{ $user->id or old('key') }}"/>
```
weerd: @DFurnes `{{ $user->id or old('key') }}` worked ;)

@deadlybutter (see above) we figured out a decent way to grab old values stored in session (for example on validation errors and you get returned to the form) or if it's already been set and you're loading the data into the field :smile:

DFurnes: Woohoo! :clap:

resources/views/users/partials/_form_users.blade.php
```diff
@@ -0,0 +1,41 @@
...
+</div>
+
+<div class="form-item -padded">
+    <label class="field-label" for="type">Select user key type:</label>
+    <div class="select">
+        <select name="type">
+            <option value="email" {{! isset($user->id) ? 'selected' : '' }}>Email</option>
```
weerd: @DFurnes Still a little iffy on dealing w/ preselected items in dropdowns and radios (below). Did this for now, but would prefer to be able to call a method, only issue is the passing of a non-set $variable...

app/Http/Controllers/UsersController.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function index()
+    {
+        $admins = User::where('role', '=', 'admin')->get();
+        $staff = User::where('role', '=', 'staff')->get();
+        $contestants = User::where('role', '=', null)->get();
```
deadlybutter: Isn't this going to get incredibly slow on a prod server without pagination?

weerd: Yeah, I need to go back in and add pagination to it and likely separate out the Admin and Staff users on one page and contestant users on another.

deadlybutter:
ah sorry forgot to :+1:

weerd:
thanks!

PR #42, opened 2016-03-03
Adds waiting room api function

### LINKED ISSUE #26: Create endpoint for phoenix to get waiting rooms
phoenix needs to ping this app to determine if there should be a competition on the node.

deadlybutter:
#### What's this PR do?
Added waiting room API response

what does this return?
`http://homestead.app/api/v1/waitingrooms?campaign_id=1`

Not using fractal because Diego is doing some fancy package magic before we start using it

Fixes #26

### DIFF
--- a/app/Http/Controllers/Api/WaitingRoomsController.php
+++ b/app/Http/Controllers/Api/WaitingRoomsController.php
@@ -8,7 +8,6 @@ use Gladiator\Models\WaitingRoom;
 
 class WaitingRoomsController extends Controller
 {
-
     /**
      * Get a waiting room for the specified campaign ID
      *
--- a/app/Http/Controllers/Api/WaitingRoomsController.php
+++ b/app/Http/Controllers/Api/WaitingRoomsController.php
@@ -15,7 +15,7 @@ class WaitingRoomsController extends Controller
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function getWaitingRoom(Request $request)
+    public function show(Request $request)
     {
         $this->validate($request, [
             'campaign_id' => 'required|integer'
--- a/app/Http/Controllers/Api/WaitingRoomsController.php
+++ b/app/Http/Controllers/Api/WaitingRoomsController.php
@@ -18,6 +18,6 @@ class WaitingRoomsController extends Controller
     {
         $waiting_room = WaitingRoom::find($request->input('campaign_id'));
 
-        return response()->json($waiting_room, 201);
+        return response()->json($waiting_room, 200);
     }
 }
--- a/app/Http/Controllers/Api/WaitingRoomsController.php
+++ b/app/Http/Controllers/Api/WaitingRoomsController.php
@@ -2,7 +2,7 @@
 
 namespace Gladiator\Http\Controllers\Api;
 
-use Illuminate\Http\Request;
+use Gladiator\Http\Requests\WaitingRoomRequest;
 use Gladiator\Http\Controllers\Controller;
 use Gladiator\Models\WaitingRoom;
 
@@ -15,14 +15,9 @@ class WaitingRoomsController extends Controller
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function show(Request $request)
+    public function show(WaitingRoomRequest $request)
     {
-        $this->validate($request, [
-            'campaign_id' => 'required|integer'
-        ]);
-
-        $campaign_id = $request->input('campaign_id');
-        $waiting_room = WaitingRoom::find($campaign_id);
+        $waiting_room = WaitingRoom::find($request->input('campaign_id'));
 
         return response()->json($waiting_room, 201);
     }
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/WaitingRoomsController.php
```diff
@@ -0,0 +1,29 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function getWaitingRoom(Request $request)
+    {
+        $this->validate($request, [
+            'campaign_id' => 'required|integer'
+        ]);
+
+        $campaign_id = $request->input('campaign_id');
+        $waiting_room = WaitingRoom::find($campaign_id);
```
weerd: no need to save this as a variable, just toss it into the method:
```
$waiting_room = WaitingRoom::find($request->input('campaign_id'));
```

app/Http/Controllers/Api/WaitingRoomsController.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Get a waiting room for the specified campaign ID
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function getWaitingRoom(Request $request)
+    {
+        $this->validate($request, [
+            'campaign_id' => 'required|integer'
```
weerd: can you set up a `WaitingRoomRequest`?

app/Http/Controllers/Api/WaitingRoomsController.php
deadlybutter: Oh I know its not required, im just used to doing this in Drupal since that seemed to be the convention there

app/Http/Controllers/Api/WaitingRoomsController.php
```diff
@@ -0,0 +1,29 @@
...
+class WaitingRoomsController extends Controller
+{
+
+    /**
+     * Get a waiting room for the specified campaign ID
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function getWaitingRoom(Request $request)
```
weerd: can we switch to Restful naming so this would just be the `show()` method?

app/Http/Controllers/Api/WaitingRoomsController.php
weerd: We're in Laravel now señor :wink:

weerd: Sounds good :smile:

app/Http/Controllers/Api/WaitingRoomsController.php
deadlybutter: yeah that makes sense

app/Http/Controllers/Api/WaitingRoomsController.php
deadlybutter: Im worried about using a WaitingRoomRequest here because it would be different than a normal request

(in this case, only campaign id is required, but in all other normal waiting room requests everything should be required)

deadlybutter:
@weerd ready for re-review!

weerd:
@deadlybutter one quick note, all api endpoints should always be plural:

```
http://homestead.app/api/v1/waitingrooms
```

weerd:
Fix Style CI issues too!

deadlybutter:
@weerd ah didnt know that, will fix that!

deadlybutter:
talked with @weerd about this in Slack --


> IMO makes more sense to just have the front end query the gladiator api, if it returns true, the frontend creates a modal
like how attn does there stuff for example
otherwise we’re still just contributing to the monolith but with less code

> if we’re moving towards a services approach, drupal should not know or care about what services its working with
JS can easily do competitions without ever touching drupal

@weerd has additional thoughts I think

cc: @angaither

deadlybutter:
@weerd fixed and fixed!

weerd:
With the new `WaitingRoomRequest` you can now remove 
```
private $validationRules = [
         'campaign_id' => 'required|numeric',
         'campaign_run_id' => 'required|numeric',
         'signup_start_date' => 'required|date',
         'signup_end_date' => 'required|date',
     ];
```
from the controller :)

app/Http/Controllers/Api/WaitingRoomsController.php
```diff
@@ -0,0 +1,23 @@
...
+     * Get a waiting room for the specified campaign ID
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function show(WaitingRoomRequest $request)
+    {
+        $waiting_room = WaitingRoom::find($request->input('campaign_id'));
+
+        return response()->json($waiting_room, 201);
```
weerd: Should this be `200`?

weerd: You're not creating a resource, just returning it:

```
201 Created
The request has been fulfilled, resulting in the creation of a new resource.[11]
```

weerd:
Just those 2 comments on items to fix and should be good!

deadlybutter:
@weerd :shipit: ?

weerd:
sorry, was out lunching! :+1:

PR #43, opened 2016-03-03
Update to fix named route path that was updated to wrong method.

weerd:
#### What's this PR do?
This PR fixes a bug routing to the wrong named route on POST for creating a new user and Laravel was throwing a `MethodNotAllowedHttpException` because it should be sending the POST request to the `store()` method.

Tyr creating a new user!

Updated to use named routes and used the wrong one :confounded: 

---


DFurnes:
:white_check_mark:

PR #44, opened 2016-03-03
User api

### LINKED ISSUE #22: Create endpoint for Phoenix to post users

weerd:
#### What's this PR do?
This PR adds ability for API POST requests to add users to the system and assign them to appropriate waiting rooms based on the `campaign_id` and `campaign_run_id` that they signed up for. It also restructures a bit so that both the Web CRUD and API can take advantage of the same methods via a new `Registrar` service (similar to how Northstar handles things) specific to the app. This allows DRYing up the code so while the routes where a User might be created might be different, the mechanisms under the hood are the same. 

It also updates the `UserRequest` validation so that depending on whether the request is coming from the web app itself or the API the validation rules can slightly change based on the requirements.

This PR also adds some initial API documentation in the repo, so it can easily be updated as endpoints get added/updated! Following the pattern @DFurnes established in Northstar :wink: 

Pull down, `composer dump-autoload`. Sign in and then create a new user. Does that work? Great! Next open up your favorite API HTTP Client and send a POST request to `http://gladiator.dev/api/v1/users` based on the documentation included in this PR and see what the response is!

@sbsmith86 based on some convos with @DFurnes I changed up how we will be moving forward using the `hasAccountInSystem()` and `hasNorthstarAccount()` methods. I left the old static ones in so it won't break stuff, but planning to remove them. Instead use the new `Registrar` class to handle all that stuff! It's cleaner and more universal now, so both the Web app and the API can easily harness those methods from their respective controllers :dancer: 

Fixes #22 

#### Next steps
- Need to create an `env` variable with a token that Phoenix will need to pass so not everyone can just sit there sending POST requests and assigning users to competitions willy-nilly.
- Fractal package has been included in project, to to apply it's use to transform the responses.

---


app/Services/Registrar.php
```diff
@@ -0,0 +1,92 @@
...
+     * @return \Gladiator\Models\User
+     */
+    public function findOrCreate($credentials)
+    {
+        $account = $this->hasAccountInSystem($credentials['type'], $credentials['key']);
+
+        if ($account instanceof User) {
+            $user = $account;
+
+            session()->flash('status', 'User already exists!');
```
DFurnes: Ideally the service class shouldn't have any presentation stuff in it (since this flash messaging will only work on web requests). It might be a good idea to break this into two separate methods like `findExistingUser` and `createUser`, so you can present the results of each step based on context (e.g. switching status code `200` to `201` for the API).

DFurnes:
Left one suggestion, but looks good to me! :car:

app/Services/Registrar.php
weerd: Yeah, I had a bit of a conflict deciding on how to resolve this... So you're saying then in the respective controllers (web vs api) it handles how to respond from there, but calls 2 methods instead of 1? So the check for `instanceof User` gets repeated in both?

documentation/readme.md
```diff
@@ -0,0 +1,4 @@
+# Gladiator API
```
angaither: :tada:

app/Services/Registrar.php
DFurnes: Yep, if you need to vary response based on things that happen in that function.

angaither:
@weerd is there anything we need to add to the .env file with this one? setting up local NS or phoenix routes? is that done already?

otherwise :+1:

weerd:
@angaither yeah, but not yet. I added a "Next steps" in the description. Gonna add the env api token along with the transformations but wanted to do in separate PR so I can do it in smaller doses and refrain from ginormous PRs :wink:

weerd:
@DFurnes @angaither I updated the code to separate concerns based on @DFurnes suggestion:

https://github.com/DoSomething/gladiator/commit/7895aaa8bd49e6f4010c71bdb6c6631b6d90cff2

angaither:
okie :+1:

DFurnes:
:+1: :+1:

PR #50, opened 2016-03-04
Adding npm install and gulp to deploy

blisteringherb:
#### What's this PR do?
This adds npm install and gulp to the Cap deploy

Deploy this branch to QA via Wercker

Nothing

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -51,17 +51,7 @@ namespace :deploy do
     run "ps -ef | grep 'queue:work' | awk '{print $2}' | xargs sudo kill -9"
   end
 
-  task :npm_install do
-    run "cd #{release_path} && npm install"
-  end
-
-  task :gulp do
-    run "cd #{release_path} && gulp"
-  end
-
 end
 
 after "deploy:update", "deploy:cleanup"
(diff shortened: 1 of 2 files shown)


DFurnes:
You shouldn’t need to do this, since Wercker uses the assets built during the `build` step for the `deploy` environment, and Capistrano deploys those. Hence why we don’t re-run `composer install` on Capistrano deploys either (or do this on any other Wercker-deployed apps).

DFurnes:
Er, well follow-up – we should add these to the Wercker build step [like we do elsewhere](https://github.com/DoSomething/aurora/blob/5441fb4906acf9686c01f7d97daaae64507bef1a/wercker.yml#L14-L23) since they're not set up in this app. That way we can build and test a fully-functioning environment & the deploy only needs to be concerned with things that would happen on an application server.

blisteringherb:
Because this app requires npm 2, I'll need to update the box that it runs on, so I'm working on that now.

blisteringherb:
I've updated the build box and the version. Gulp and npm install seem to be working now. @angaither

angaither:
:+1:

Issue #45, opened 2016-03-03
update db structure 


angaither:
New DB structure. 

**Overview:** A `Contest` is the umbrella container for everything. Each `Contest` has one `Waiting Room` that has sign up dates tied to it that can be set and edited by an admin as needed. `Waiting Rooms` get split into `Competitions`. Competitions hold groups of users competing in a contest and store the date the competition starts and the date they will end based on the run period of the `Contest`.

#### Contests
* `contest_id`
* `campaign_id` 
* `campaign_run_id`
* `waiting_room_id`
* How long the contest runs for (`duration`)

Also has (currently undetermined) metadata such as: 
* title 
* description
* prize (?)
* rules

#### Waiting Rooms
* `waiting_room_id`
* `contest_id`
* `signup_start_date` (admin can edit)
* `signup_end_date` (admin can edit)

#### Waiting Room Users
NOTE: I think this table exists already and should still work with this new structure:
`waiting_room_id`
`user_id`

#### Competitions
* `competition_id`
* `contest_id`
* `competition_start_date` (based on the day this competition was created)
* `competition_end_date` (calculated based on `run_period` and set on creation

#### Competition Users
* `competition_id`
* `user_id`

angaither:
from @sbsmith86 
>  there will be a Brackets table that can have a model that defines all the methods we need to work with brackets. For example, per competition, we might want to show all the brackets, and then under each bracket, all of the users in that bracket.

weerd:
#### Notes

Some thoughts to follow up on our #Slack conversation from last friday:

- Competition table holds all the metadata for a competition, stuff that maybe Phoenix would want to know of when setting up a competition signup popup, such as:
  - title
  - campaign_id
  - campaign_run_id
  - signup_copy
  - signup_button_text
  - confirmation_message
  - etc.
- Competitions could also hold the full timeline for a particular competition based on a series of dates:
  - signup_start (signup period to know if competition is active)
  - signup_end (signup period to know if competition is active)
  - competition_start
  - competition_end
- WaitingRooms are created for a specific Competition and should be considered a "contestant limbo" that temporary holds contestants until they are broken up into brackets at start of the next iteration of a particular competition.
- There can be some logic added that creates WaitingRooms for a Competition as the prior WaitingRoom closes and splits it's group of users. These WaitingRooms could be created up to the end of the Competition `signup_end` which would signal no more WaitingRooms cause we're on the last round of a Competition.

deadlybutter:
[...]

Issue #27, opened 2016-02-29
Display users info in competitions


angaither:
users in a competition should be displayed with their reportback activity

PR #51, opened 2016-03-04
Show competition users

sbsmith86:
#### What's this PR do?

Shows a table of users in a competition. 

I did a bunch of things in the PR -

* Hooked up to Phoenix so that we can pull user activity. I am holding off on showing actual signup activity for now because this was getting hard without any real phoenix users in the system. 
* Made a new endpoint `/users/{id}/signup/{signup_id}` that can show the user activity. TODO: add campaign run as a parameter so you can see only signup/reportback activity for the campaign and campaign run a user signed up for.
* Added a function in the Competitions model that gets all of the users in a competiton. I wasn't sure where the correct place to put a function like this was. Ideally, when we finish #45, there will be a Brackets table that can have a model that defines all the methods we need to work with brackets. For example, per competition, we might want to show all the brackets, and then under each bracket, all of the users in that bracket. 
* Added a seeder to create a dummy competition in my DB. I didn't add this to the root seeder class cause only wanted to run it, if I needed it, but it's there. 
* Added a `competition_users` seeded that put a random set of users in a competition so I can work with them. This will probably be depreciated when we move to the Bracket structure.
* Lastly, a small test that just checks that I am hitting and getting a response from phoenix. This is not a great test, but it was easier for me to do this then to do like a `dd()` and have to refresh my browser everytime. 

pull this branch, run the seeds, go to view a competition, and you should see: 

screen shot 2016-03-04 at 2 20 22 pm

This is still a work in progress, but gets us mostly there. 

There is more to be done re: showing user signup data, which can be done easier now since we have db seeds that have real phoenix users. 

The test needs to be fixed to use mock users.

Address #27 

cc @weerd @angaither


config/services.php
```diff
@@ -29,6 +29,11 @@
         'secret' => env('MANDRILL_SECRET'),
     ],
 
+    'phoenix' => [
+        'uri' => env('PHOENIX_URI'),
```
angaither: it'd be nice to add default values for this in `env.example`

sbsmith86: @angaither ah like the staging url?

angaither:
@sbsmith86 doesn't `view signup` also display rb data?

weerd:
> For example, per competition, we might want to show all the brackets, and then under each bracket, all of the users in that bracket.

Really like this idea.

app/Http/Controllers/CompetitionsController.php
```diff
@@ -59,7 +60,13 @@ public function store(CompetitionRequest $request)
      */
     public function show(Competition $competition)
     {
-        return view('competitions.show', compact('competition'));
+        $bracket = Competition::getBracket($competition->id);
```
weerd: This isn't implement yet right, cause we don't have Brackets yet?

Would this be plural though, since multiple brackets could apply to the `$competition->id`?

angaither:
if https://github.com/DoSomething/gladiator/pull/51#issuecomment-192438416 is true, it might be better to call it 'view activity'

app/Http/Controllers/CompetitionsController.php
sbsmith86: @weerd it's "implemented" in the since that all it does is get all users from the `competition_users` table that have that competition id...but that needs to be updated once we have brackets. and :+1: to the pluralization.

sbsmith86:
@angaither that makes sense

app/Models/Competition.php
```diff
@@ -15,4 +16,18 @@ public function users()
...
+     *
+     * @TODO - right now this just gets all the users in the competitions_user table that have the correct competition id
+     * however, I imagine, we would want some sort of "bracket id", so we can list users by their bracket instead of one,
+     * all encompassing list.
+     *
+     * Once the Bracket table is set up, we should have a bracket model, that defines a query scope to get all users within a bracket.
+     */
+    public static function getBracket($competition_id)
+    {
+        return DB::table('competition_user')->where('competition_id', $competition_id)->get();
```
weerd: Maybe I'm confused cause we're not dealing w/ Brackets officially yet, but why are you hitting the pivot table for the record? Are you grabbing the Users(?) where the competition_id is `$competition_id`?

sbsmith86: @weerd yes, currently that seem the only way to get get all the user in a competition. But onces brackets are implemented all this will change. Maybe I shouldn't have named it Brackets. But i just liked the name lol

sbsmith86: my TODO comment is awkward. I'm just trying to say we need the bracket table set up to do this correctly.

.env.example
```diff
@@ -24,3 +24,5 @@ MAIL_ENCRYPTION=null
 
 NORTHSTAR_URI=http://northstar-qa.dosomething.org
 NORTHSTAR_API_KEY=abc4324
+
+PHOENIX_URI=https://staging.beta.dosomething.org/api
```
angaither: version too!

angaither:
looks :ok: from here

app/Models/Competition.php
weerd: Gotcha!

app/Services/Phoenix/Phoenix.php
```diff
@@ -0,0 +1,41 @@
...
+namespace Gladiator\Services\Phoenix;
+
+use Gladiator\Services\RestApiClient;
+
+class Phoenix extends RestApiClient
+{
+    protected $base_uri;
+
+    /**
+     * Northstar constructor.
```
weerd: comment typo!

app/Models/User.php
```diff
@@ -99,4 +99,36 @@ public static function hasNorthstarAccount($type, $id)
 
         return $northstar->getUser($type, $id);
     }
+
+    /**
+     * Create an array with only the useful user information we need for use in views, etc.
+     */
+    public static function setUserInfo($user)
```
weerd: We can eventually take this out and put to another more accessible class that both the web and api can use in their respective controller :)

weerd:
Post-merge :+1:

app/Models/User.php
sbsmith86: @weerd yup! cool

app/Services/Phoenix/Phoenix.php
sbsmith86: @weerd ack

Issue #33, opened 2016-02-29
seed with users from NS


angaither:
from staging 

store that info in cache

PR #52, opened 2016-03-04
Add hils and charbur to seedery.

angaither:
#### What's this PR do?
adding more people as admins on seeding

again run  `php artisan db:seed --class=UserTableSeeder`

Refs #33


weerd:
:+1:

PR #55, opened 2016-03-07
Transforming like a boss

### LINKED ISSUE #48: Implement API Transformers & API Key
Transformers, robots in disguise!
image

Implement the use of API response Transformers using Fractal and add an API key so POST can't be sent to Gladiator anonymously!

Relates to recent work completed in #44

weerd:
#### What's this PR do?
This PR implements the use of the Fractal library to transform API responses and standardize what they should be across the Gladiator API. It also updates the `/api/v1/users` POST endpoint to take into account the new DB schema and checks against the Contest -> WaitingRoom to see if a user already exits in the room before adding them.

This also removes the `/api/v1/waitingrooms` endpoint which we are no longer using with the new DB schema, and replaced it with `/api/v1/contests`.

Pull down branch, `composer dump-autoload` try hitting `/api/v1/users` with a POST request based on the [API doc](https://github.com/DoSomething/gladiator/blob/master/documentation/endpoints/users.md#create-a-user) and you should see a JSON response with the user info (keep in mind all values apart from `id` will be null cause that shiz is coming in a different PR with caching the user data!)

```javascript
{
  "data": {
    "id": "697efkhad04858403adnas43",  // MongoDB id
    "first_name": null,
    "last_name": null,
    "email": null,
    "mobile": null,
    "signup": null,
    "reportback": null
  }
}
```

Waiting on Contest CRUD #59 before I can format and return an updated response for Phoenix when asking about contest availability.

Fixes #48 

---
@DFurnes @angaither 

cc: @sbsmith86 @deadlybutter


app/Http/Controllers/Api/ApiController.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    protected $transformer;
+
+    /**
+     * Create new ApiController instance.
+     */
+    public function __construct()
+    {
+        $this->manager = new Manager;
+        $this->manager->setSerializer(new DataArraySerializer);
```
weerd: I'm thinking it might be best to pull these out of the `__construct()` and place in the `transformer()` method so that the controller extending this ApiController doesn't have to always declare `parent::__construct()` in its `__construct()`...

DFurnes: Yeah, I think that's a good idea.

app/Http/Controllers/Api/ApiController.php
```diff
@@ -0,0 +1,89 @@
...
+     * @param  object  $data
+     * @param  int  $code
+     * @param  array  $meta
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function transform($data, $code = 200, $meta = [])
+    {
+        $data->setMeta($meta);
+
+        $this->manager = new Manager;
```
DFurnes: Is the manager used outside this method? Might not need it to be a property on the class if not.

app/Http/Controllers/Api/ApiController.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $item = new Item($data, $this->setTransformer($transformer));
+
+        return $this->transform($item, $code, $meta);
+    }
+
+    /**
+     * Manage and finalize the data transformation.
+     *
+     * @param  object  $data
```
DFurnes: Since this only accepts an `Item` or `Collection`, it'd be nice to specify here.

app/Http/Controllers/Api/ContestsController.php
```diff
@@ -0,0 +1,31 @@
...
+     * @var \Gladiator\Http\Transformers\UserTransformer
+     */
+    protected $transformer;
+
+    /**
+     * Create a new ContestsController instance.
+     *
+     * @param \Gladiator\Http\Transformers\ContestTransformer $transformer
+     */
+    public function __construct(ContestTransformer $transformer)
```
DFurnes: I don't know if there's much value in injecting this from the service container (rather than setting `$this->transformer = new ContestTransformer`) since it doesn't have any dependencies of it's own.

app/Http/Controllers/Api/UsersController.php
```diff
@@ -44,20 +49,20 @@ public function store(UserRequest $request)
...
+            return $this->item($user);
         }
 
-        $waitingRoom->users()->attach($user->id);
-
-        // @TODO: return Transformed response via Fractal
-        return 'User was not in waiting room, so they have been assigned to it!';
+        $contest->waitingRoom->users()->attach($user->id);
+            // @TODO: maybe add more detail to response to indicate which room user was added to?
+            return $this->item($user);
```
DFurnes: Looks like indentation is a bit off here. :space_invader:

app/Http/Controllers/Api/ApiController.php
weerd: oh good call. Yeah, it's not, and if it is eventually I could just move it back out.

app/Http/Controllers/Api/ApiController.php
weerd: ¡Si!

DFurnes:
Left some nitpicks, but looks good to me! :ship: :shipit:

DFurnes:
I'd also love to pull the `ApiController` class here into a trait and publish it as a package that any of our Laravel APIs can use (so we don't have to set this up by hand each time).

app/Http/Controllers/Api/ContestsController.php
weerd: so same same for the `Api/UsersController` too?

weerd:
Si! Maybe post Gladiator v1 we can pull `ApiController` out?

app/Http/Controllers/Api/ContestsController.php
DFurnes: Yeaaaa, I kinda prefer that. I just checked in Northstar and we do it both ways though. :joy:

DFurnes:
Yeah, definitely not a high priority thing... just noting for later. :smile:

PR #56, opened 2016-03-08
DB Restructure

### LINKED ISSUE #45: update db structure

sbsmith86:
#### What's this PR do?

Updates the DB structure in Gladiator as outlined in #45 

* Removed old migrations, and made new ones. 
* Added a new `Contests` table and a model that defined it's relationship to waiting rooms and competitions
* Renamed pivot tables so they get run last. 
* Added a new `ContestTableSeeder` that makes 3 contests.
* Added a `ContestServiceProvider` that listens for the `created` event on Contest models and automatically creates a new waiting room for the competition. 
* Created a competition seeder that just creates three empty competitions for a Contest.

NOTES/QUESTIONS -

* I wasn't sure about the `duration` field. My thought is we can make the decision to always store this as an `integer` representing days and then on the front end we can provide a slicker interface with like units (days, weeks, months, etc) that gets converted into days before it gets stored in the db. But I am open to doing something different. 

* Because of the ^^ open question I didn't implement any fancy calculation of competition start and end date yet. 

* I didn't touch any CRUD stuff or other logic to keep this as small as I could, so we can fix those up in other issues. 

Pull down this branch
Run `composer dump-autoload`
Run `php artisan:refresh` to get new tables
Run `php artisan db:seed` to seed the DB

Review the DB to make sure all the correct columns are there and the relationships look right.

Fixes #45 

@angaither @deadlybutter @weerd


angaither:
my big feeling on `duration` on contests is that it probably shouldn't be there. 
we have dates on the waiting room and that should be accessible to the competition but shouldn't be a different value since it's just so many dates.

sbsmith86:
@angaither The waiting room dates designate the length of the sign up period and the competition duration is supposed to help us track how long a competition will run for (so like how long each group of people will actually have to compete) How would we do that without `duration`?

angaither:
@sbsmith86 ahh, my bad. I thought `duration` on the `contest` was how long the signup period should be open. totally misunderstood that field

app/Providers/ContestServiceProvider.php
```diff
@@ -0,0 +1,42 @@
...
+        Contest::created(function ($contest) {
+            $waitingRoom = new WaitingRoom;
+
+            $waitingRoom->contest_id = $contest->getKey();
+
+            // Default sign up period of one week.
+            $waitingRoom->signup_start_date = Carbon::now();
+            $waitingRoom->signup_end_date = Carbon::now()->addWeeks(1)->endOfDay();
+
+            $waitingRoom->save();
```
weerd: Nice!

weerd:
> @angaither The waiting room dates designate the length of the sign up period and the competition duration is supposed to help us track how long a competition will run for (so like how long each group of people will actually have to compete) How would we do that without duration?

@sbsmith86 Is it the competition `duration` or the contest `duration`?

sbsmith86:
@weerd ah yea sorry. it's the Contest `duration`...new terminology got me so confused lol. I'm really working on it. One step at a time.

angaither:
now I'm confused again,
i thought it was the `competition` duration that's the first piece of metadata on the contest that's inherited on the all the `competitions`

sbsmith86:
@weerd @angaither it is the `duration` field on the `contests` table that will be used to set `competition_start_date` and `competition_end_date` for each competition in a contest. This way, competitions can start at different times (based on when they were created) but all run for the same amount of time.

weerd:
No worries. So yeah, if it's the contest `duration`, I'm still not sure why we couldn't have dates for that: `contest_start` & `contest_end`, which otherwise match the campaign. A `duration` could easily be calculated from the start and end dates. I think I'm confused too though.

If it's the competition `duration` then that would dictate how long a single competition in the contest has to run, with each competition in a contest starting at different points; but they all run for say a duration of 10 days...

weerd:
Ok, so maybe what's confusing is this comment:

> `$table->integer('duration')->comment = 'Duration of the campaign in days';`

campaign -> competition?

weerd:
> @weerd @angaither it is the duration field on the contests table that will be used to set competition_start_date and competition_end_date for each competition in a contest. This way, competitions can start at different times (based on when they were created) but all run for the same amount of time.

This makes sense; just change up the comment added to the `duration` column in the migration ;)

sbsmith86:
Sorry for the confusion. 

* So an admin creates a contest and says each competition within that contest should run for 1 month (`duration`)
* A waiting room is created and the admin can edit the start and end dates of the waiting room, so lets say they want to collect signups for two weeks. 
* Two weeks go by and the admin is ready to break people out into competitions. 
* They split the waiting room
* The competitions are created and gladiator uses the `duration` to determine the start and end date for each competition group. 
* BUT based on the amount of people, we only split into two competitions and the admin wants to do more signups. 
* they update the waiting room period to new dates
* that signup period is over, and they split the waiting room
* the `duration` field is then used again to figure out the start and end dates for the competitions that were just created.

deadlybutter:
^ what @sbsmith86 just said, :+1:

sbsmith86:
@weerd My bad! yea that comment is :hankey:

weerd:
@sbsmith86 excellent example scenario! very helpful! 

looks good :+1:

angaither:
:+1: 
nice job on this.

PR #58, opened 2016-03-08
Use many to many eloquent relationship to save competitions

sbsmith86:
#### What's this PR do?

I update the Contest relationship to competitions to be `hasMany()` in the model and so I updated the method to be plural. However, I forgot to use the plural version of the function when I was calling it in the competition seeder. 

Tested on my local and we should be good. you can use the same steps as in #56 

Ref #45


angaither:
:+1:

PR #64, opened 2016-03-09
Update waiting rooms crud

### LINKED ISSUE #60: Update Waiting Rooms CRUD
Update the waiting room CRUD methods and views to work with the new DB structure. 

After #56, we are now creating a waiting room automatically when a contest is created. We still need the option for admins to be able to view all waiting rooms, edit waiting rooms, etc. 

I'm not sure we need to expose the create form anymore but open to thoughts on that.

sbsmith86:
#### What's this PR do?
* Removed the ability to create and destroy waiting rooms since these get created automatically when a contest gets created, and we also don't want to be deleting waiting rooms since a contest should never not have a waiting room. 

* Updated index and show views to work with the new waiting rooms table. 

* Updated the edit form and validation rules to work with the new waiting rooms table. 

Pull down this branch and check the CRUD functionality or should I say the RU functionality :smirk: 

Make sure it still works with the new DB structure. 

Making these updates after #45 was done.

Fixes #60 


resources/views/waitingrooms/show.blade.php
```diff
@@ -10,8 +10,7 @@
     <div class="container">
         <div class="wrapper">
             <div class="container__block">
-                <p>Campaign ID: {{ $room->campaign_id }}</p>
```
angaither: i know these aren't on the room, but it might be helpful to show it from the contest

resources/views/waitingrooms/index.blade.php
```diff
@@ -9,32 +9,24 @@
...
                       </thead>
                       <tbody>
                         @foreach($rooms as $room)
                           <tr class="table__row">
                             <td class="table__cell"><a href="{{ route('waitingrooms.show', $room->id) }}">{{ $room->id }}</a></td>
-                            <td class="table__cell">{{ $room->campaign_id }}</td>
-                            <td class="table__cell">{{ $room->campaign_run_id }}</td>
+                            {{-- TODO: Link to contest. --}}
+                            <td class="table__cell"><a href="#">{{ $room->contest_id }}</a></td>
                             <td class="table__cell">{{ date('F d, Y', strtotime($room->signup_start_date)) }}</td>
```
DFurnes: This should be returned as a Carbon instance from the model (I thiiiiiink) so you should be able to do `$room->signup_start_date->format('F d, Y');` rather than converting it to a string and formatting that.

resources/views/waitingrooms/show.blade.php
sbsmith86: @angaither ok cool. Also, i think ideally we should show more user friendly info like Campaign title or something, but maybe we can do that in a different issue. I don't want to add too much to this.

resources/views/waitingrooms/index.blade.php
sbsmith86: :+1:

resources/views/waitingrooms/show.blade.php
angaither: totes

sbsmith86:
@angaither @DFurnes Can I merge this?

DFurnes:
Yeah, looks good to me! :shipit:

weerd:
sorry! was out lunching! looks good :+1:

Issue #57, opened 2016-03-08
Add more info to competition user view


angaither:
- tables should be sortable by various things
- expose rb quantity 
- updated_at

sbsmith86:
is this a dupe of #27

angaither:
yup, i added more info here though, closing #27

PR #68, opened 2016-03-09
Update competitions crud

### LINKED ISSUE #61: Update Competitions CRUD
Update the competition CRUD methods and views to work with the new DB structure as needed.

sbsmith86:
#### What's this PR do?

Update the competition CRUD stuff to work with the new DB structure. 

Pull down this branch
Run `composer dump-autoload` 
Create a competition, edit a competition, delete a competition, view all competitions. 

I removed the weird bracket stuff i had in there an we can re-address showing users and user activity in a competition in #57 

Fixes #61 

cc @DFurnes @weerd @deadlybutter @angaither

### DIFF
--- a/app/Models/Competition.php
+++ b/app/Models/Competition.php
@@ -19,7 +19,7 @@ class Competition extends Model
      *
      * @var array
      */
-    protected $dates = ['start_date', 'end_date'];
+    protected $dates = ['competition_start_date', 'competition_end_date'];
 
     /**
      * A Competition belongs to many Users.
--- a/app/Models/Competition.php
+++ b/app/Models/Competition.php
@@ -2,7 +2,6 @@
 
 namespace Gladiator\Models;
 
-use DB;
 use Illuminate\Database\Eloquent\Model;
 
 class Competition extends Model
--- a/app/Models/Competition.php
+++ b/app/Models/Competition.php
@@ -28,18 +28,4 @@ class Competition extends Model
     {
         $this->belongsToMany(User::class);
     }
-
-    /**
-     * Get all users in a competition.
-     *
-     * @TODO - right now this just gets all the users in the competitions_user table that have the correct competition id
-     * however, I imagine, we would want some sort of "bracket id", so we can list users by their bracket instead of one,
-     * all encompassing list.
-     *
-     * Once the Bracket table is set up, we should have a bracket model, that defines a query scope to get all users within a bracket.
-     */
-    public static function getBrackets($competition_id)
-    {
-        return DB::table('competition_user')->where('competition_id', $competition_id)->get();
-    }
 }
--- a/app/Models/Competition.php
+++ b/app/Models/Competition.php
@@ -12,7 +12,7 @@ class Competition extends Model
      *
      * @var array
      */
-    protected $fillable = ['campaign_id', 'campaign_run_id', 'start_date', 'end_date'];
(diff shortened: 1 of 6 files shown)


resources/views/competitions/partials/_form_competitions.blade.php
```diff
@@ -1,23 +1,18 @@
 @include('layouts.errors')
 
 <div class="form-item -padded">
-    {!! Form::label('campaign_id', 'Campaign ID:', ['class' => 'field-label']) !!}
-    {!! Form::text('campaign_id', (isset($room->campaign_id)) ? $room->campaign_id : NULL, ['class' => 'text-field']) !!}
+    {!! Form::label('contest_id', 'Contest ID:', ['class' => 'field-label']) !!}
+    {!! Form::text('contest_id', (isset($competition->contest_id)) ? $competition->contest_id : NULL, ['class' => 'text-field']) !!}
```
weerd: I mentioned this in @deadlybutter PR too, but can you update this to use:

```php
{{ $competition->contest_id or old('contest_id') }}
```
It'll help with repopulating the field if there was a validation error with what the user entered and also feels much cleaner :) It's a little useful syntactic sugar :)

app/Models/Competition.php
```diff
@@ -12,14 +11,14 @@ class Competition extends Model
      *
      * @var array
      */
-    protected $fillable = ['campaign_id', 'campaign_run_id', 'start_date', 'end_date'];
+    protected $fillable = ['contest_id', 'competition_start_date', 'competition_end_date'];
```
angaither: should `contest_id` be fillable? isn't it auto increment-able?

sbsmith86: A competition just references a the contest that it is apart of. This is fillable so we can create competitions and set the contest_id for them.

weerd: Would the system autofill it at all? Given it's creating a competition for a specified contest? Or do we need to set the contest id each time a competition is created, which is fine too, just more work...

resources/views/competitions/partials/_form_competitions.blade.php
sbsmith86: @weerd this doesn't seem to work on these date fields when setting the default value. Am I doing something wrong here: 

```{!! Form::date('competition_start_date', $competition->competition_start_date or old($competition->competition_start_date), ['class' => 'text-field']) !!}```

app/Models/Competition.php
deadlybutter: in theory its just 1 line of code when creating the competitions ($competition->contest_id = $contest->id)

I think its fine being fillable

sbsmith86: I believe, yes, gladiator creates the competitions automatically when the waiting rooms get split ( @deadlybutter is that correct?) But even so, doesn't it need to be able to be mass assigned?

deadlybutter: yeah thats how its supposed to work, i think its broken right now cause of all the DB updates

sbsmith86: How are we feeling about this.

deadlybutter: Im fine with it, @weerd @angaither ?

angaither:
:+1:

app/Models/Competition.php
weerd: Not a huge deal. I think it's fine. If we eventually decide against it, it's an easy fix as @deadlybutter pointed out. Plus it's not the actual contest id being assigned, so should be fine.

PR #71, opened 2016-03-10
Updates split functionality based on contests

### LINKED ISSUE #62: Update Waiting Room Split functionality
Update the waiting room split functionality to work with the new DB structure, not sure how much of this needs to change, but this is a general outline of what should happen. 

* There is one waiting room per contest.
* When the sign up period is over there is an option to split the room. 
* The room gets split in to competitions.
* After split, the users get removed from that room and it can be reused if needed.

deadlybutter:
#### What's this PR do?
Updates the split functionality to use the new contest-y way of doing things
Removes the competition form on the split page (don't need it anymore!)
Instead it just uses the contest ID and does some simple duration adding / carbon calculations

So I updated the UserTableSeeder with this 

```
        $seedContestants = [];

        for ($i = 1; $i < 10; $i++) {
            $seedContestants = array_merge($seedContestants, $this->northstar->get('users', ['limit' => 100, 'page' => $i]));
        }
```

which got 900 users 

I also changed this bit in the contestant loop

```
            $index = mt_rand(0, ($totalRooms - 1));
            $index = 0;
```

so I could test it split 900 people in a waiting room correctly.

This is a screenshot of the /split page

screen shot 2016-03-10 at 2 16 55 pm

Then this is after the DB split.

screen shot 2016-03-10 at 2 17 38 pm

Also the competitions index

screen shot 2016-03-10 at 2 24 26 pm

Fixes #62

### DIFF
--- a/app/Models/WaitingRoom.php
+++ b/app/Models/WaitingRoom.php
@@ -2,6 +2,7 @@
 
 namespace Gladiator\Models;
 
+use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Model;
 
 class WaitingRoom extends Model
@@ -34,6 +35,7 @@ class WaitingRoom extends Model
      */
     public function getDefaultSplit()
     {
+
         $users = $this->users;
 
         // Get the size of the waiting room
@@ -66,11 +68,20 @@ class WaitingRoom extends Model
     /*
      * Creates competitions based on the given user split.
      */
-    public function saveSplit($competitionInput, $split)
+    public function saveSplit($contest, $split)
     {
+        $startDate = Carbon::now()->startOfDay();
+        $endDate = Carbon::now()->addDays($contest->duration);
+
         foreach ($split as $competitionGroup) {
             // For each split, create a competition.
-            $competition = Competition::create($competitionInput);
+            $competition = new Competition;
+
+            $competition->contest_id = $contest->getKey();
+            $competition->competition_start_date = $startDate;
+            $competition->competition_end_date = $endDate;
+
+            $contest->competitions()->save($competition);
 
             // For each user in this group
             foreach ($competitionGroup as $userId) {
--- a/app/Models/WaitingRoom.php
+++ b/app/Models/WaitingRoom.php
@@ -35,7 +35,6 @@ class WaitingRoom extends Model
      */
     public function getDefaultSplit()
     {
-
         $users = $this->users;
 
         // Get the size of the waiting room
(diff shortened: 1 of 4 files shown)


app/Models/WaitingRoom.php
```diff
@@ -66,11 +67,20 @@ public function getDefaultSplit()
     /*
      * Creates competitions based on the given user split.
      */
-    public function saveSplit($competitionInput, $split)
+    public function saveSplit($contest, $split)
     {
+        $startDate = Carbon::now()->startOfDay();
+        $endDate = Carbon::now()->addDays($contest->duration);
```
sbsmith86: probably should set this to the end of the day with `->endOfDay()`

database/seeds/UserTableSeeder.php
```diff
@@ -50,10 +50,21 @@ public function run()
...
+
+        $seedContestants = [];
+
+        for ($i = 1; $i < 10; $i++) {
+            $seedContestants = array_merge($seedContestants, $this->northstar->get('users', ['limit' => 100, 'page' => $i]));
+        }
 
         foreach ($seedContestants as $contestant) {
             $index = mt_rand(0, ($totalRooms - 1));
+            $index = 0;
```
angaither: why is index always 0?

app/Models/WaitingRoom.php
deadlybutter: ah yeah good idea

database/seeds/UserTableSeeder.php
sbsmith86: Why set this to zero after it gets set it to a random index before this?

deadlybutter: apparently i forgot to remove all of my test code in the seed file

database/seeds/UserTableSeeder.php
```diff
@@ -50,10 +50,21 @@ public function run()
         // Add Contestant Users
         $waitingRooms = WaitingRoom::all();
         $totalRooms = count($waitingRooms);
-        $seedContestants = $this->northstar->getSeedUsers(300);
+        // $seedContestants = $this->northstar->getSeedUsers(300);
```
weerd: Actually, we should probs just update that method in `Northstar.php` to be `getUsers($limit = '100', $page = 1)` or whatevs so we have that as a method in there not just for seeding.

deadlybutter: please see what i put in slack, just ignore these changes

sbsmith86:
looks good to me

PR #72, opened 2016-03-10
Updating composer packages.

angaither:
#### What's this PR do?
updates some stuff in `composer.json`

run `composer dump-autoload`

I kept getting this error below, and composer update fixed it  `¯\_(ツ)_/¯ `
```
FatalErrorException in ClassLoader.php line 347:
Maximum function nesting level of '100' reached, aborting!
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -225,16 +225,16 @@
         },
         {
             "name": "guzzlehttp/promises",
-            "version": "1.0.3",
+            "version": "1.1.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/guzzle/promises.git",
-                "reference": "b1e1c0d55f8083c71eda2c28c12a228d708294ea"
+                "reference": "bb9024c526b22f3fe6ae55a561fd70653d470aa8"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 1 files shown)


sbsmith86:
the only updates are to `composer.lock`? Not `composer.json`?

weerd:
https://laracasts.com/discuss/channels/general-discussion/l5-maximum-function-nesting-level-of-100-reached-aborting

angaither:
@sbsmith86 oh, i guess?

weerd:
If the `composer.json` doesn't specify an exact version but allows minor updates then that'll happen and only update the `composer.lock`.

sbsmith86:
@weerd ah i see, i guess i don't understand why it's in version control. Maybe we all just need to run `composer update`?

angaither:
@sbsmith86 for the [bible tells us so](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file)

sbsmith86:
@angaither ahhh gotcha!

angaither:
can i get a thumbs up here or wat

sbsmith86:
:tada:

PR #73, opened 2016-03-10
removes create competition button

deadlybutter:
#### What's this PR do?
removes the create competition button

is the button there?

thanks fantini for pointing this out

Fixes #no issue for this


deadlybutter:
that sounds like a good idea

deadlybutter:
removed! :tada:

angaither:
yeah, but also the route, I think you have to say `except ['create']` 
you can still go there, and the route will fail

deadlybutter:
ah ok

sbsmith86:
:+1:

angaither:
I don't know if that will mess up, if it's automatically getting created?

deadlybutter:
@angaither why would anyone visit that route? i tried visiting locally and its just an error page saying it doesnt exist

angaither:
@deadlybutter they shouldn't be, right?
they get created when you split the waiting room, so yeah, my first comment was right, you should add the `except` in

deadlybutter:
@angaither gotcha, added it to the routes file

angaither:
:+1:

PR #74, opened 2016-03-11
Caching of User data from Northstar.

### LINKED ISSUE #53: implement user cache
store user info in cache 
- first name
- last name
- email
- mobile

this part may be vNext
- signup and rb data

weerd:
#### What's this PR do?
This PR adds a caching layer within a system for User Repository data retrieval. There is now a dedicated user repository for when user data is fetched. The basic gist is that the system always checks the cache first and if it doesn't find the data it needs resorts to doing a full lookup.

There were a ton of little "gotchas" and other challenges, so feel free to ask me lots of questions! :dancer: 

Pull this shiz down, `composer dump-autoload`, then log in, try adding some users, changing their roles, and seeing if all updates as expected when viewing them again in the index listing or on their individual show pages.

Instead of caching specific queries and entire collections of users which would repeat a lot of the data in the cache, the system caches each user individually, keyed by their Northstar ID. So when a single request to find a specific user is done, the `CacheUserRepository` first checks **Redis** for that user via their specified id, if none is found then it passes request off to the `DatabaseUserRepository`where it will find the user in the database, and use their ID to get their full profile data from Northstar. The returned data from Northstar is then passed back to the `CacheUserRepository`, where it is properly cached and then returned.

The extra magic :tophat: :sparkles: happens when a request is made to find a collection of users. Instead of a cached collection of that specific request, it firsts checks Redis for a a cache of the requested IDs, and then uses those IDs to find the individual users in cache and returns them as a collection, harnessing some of the multiple store/retrieve methods available in Laravel's `Cache` class. So that single user found in cache before, is the same record from cache returned in the collection. :boom: 

_[Interesting fact](http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers)..._ URLs have a character limit (approx 2000). When using mongo IDs (each has 24 characters) that are passed as comma separated values of users for Northstar to search for, you can break the browser and make an invalid request when you combine 298 (which is what I was trying for all the seeded contestants, but ~83 is the actual max limit) or so ids into a single string and pop them in URL as part of the request :scream: So I ended up having to batch process requests to Northstar in 50 user increments!

Fixes #53

#### Shoutouts
To Mr. @DFurnes who helped me think through of a few of these issues and provided some useful guidance! :trophy: 

---


app/Http/Controllers/UsersController.php
```diff
@@ -33,11 +42,10 @@ public function __construct(Registrar $registrar)
      */
     public function index()
     {
-        $admins = User::where('role', '=', 'admin')->get();
-        $staff = User::where('role', '=', 'staff')->get();
-        $contestants = User::where('role', '=', null)->get();
+        $users = $this->user->getAll();
+        // dd($users);
```
deadlybutter: oops!

app/Repositories/CacheUserRepository.php
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Get collection of users from cache by specified role or default to database lookup.
+     *
+     * @param  string $role
+     * @return \Illuminate\Support\Collection
+     */
+    public function getAllByRole($role)
+    {
+        $key = $role ? $role : 'contestant';
```
DFurnes: Could this be a default parameter for the function?

resources/views/users/index.blade.php
```diff
@@ -14,15 +14,15 @@
...
         @include('users.partials._table_users', ['users' => $admins, 'role' => 'Admins'])
     @endif
 
-    @if (count($staff) > 0)
+    @if ($staff->count())
         @include('users.partials._table_users', ['users' => $staff, 'role' => 'Staff'])
     @endif
 
-    @if (count($contestants) > 0)
+    @if ($contestants->count())
```
DFurnes: Would this make more sense as `@if (! empty($contestants))`?

app/Repositories/CacheUserRepository.php
weerd: Oh yeah, good call :smile:

app/Repositories/UserRepositoryInterface.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Gladiator\Repositories;
+
+interface UserRepositoryInterface
```
DFurnes: This is such a nitpick, but Taylor seems to like calling these "Contracts" in the Laravel world.

resources/views/users/index.blade.php
weerd: it's a collection and the collection technically contains an `items` property that is an empty array (if nothing inside), so I figured I'd use the `count()` method available on the collection, cause I was afraid anything else might do a false positive.

app/Repositories/UserRepositoryInterface.php
weerd: tableflip! ok, I can update! standards, standards, standards!

app/Repositories/CacheUserRepository.php
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * Remove an item from the cache.
+     *
+     * @param  string  $key
+     * @return void
+     * @TODO: Might be best to return a bool like the Laravel class does?
+     */
+    protected function forget($key)
```
DFurnes: These feel a little bit like premature optimization, since they're just one-line aliases to framework methods... I could be wrong, but let's just keep an eye on whether or not they seem useful in the future.

app/Repositories/CacheUserRepository.php
```diff
@@ -0,0 +1,231 @@
...
+use Illuminate\Support\Facades\Cache;
+
+class CacheUserRepository implements UserRepositoryInterface
+{
+    /**
+     * UserRepositoryInterface instance.
+     *
+     * @var \Gladiator\Repositories\UserRepositoryInterface
+     */
+    protected $database;
```
DFurnes: Another nitpick, but this should probably just be `protected $repository` since in theory you could nest any "conforming" repository in here that matches the contract.

app/Repositories/CacheUserRepository.php
weerd: Yep yep. I think originally I had a few more things happening in my aliased method, but then removed them. But yeah, let's keep :eyes:

deadlybutter:
@weerd that is :indeed: an interesting fact in your context thing

app/Repositories/CacheUserRepository.php
weerd: Yeah, I was just afraid it wouldn't be as straightforward since you're _in_ the Cache Repository and then requesting from another repository might seem like requesting from the same one :confounded: I can update if it doesn't feel right though!

angaither:
a quick pr

resources/views/users/index.blade.php
DFurnes: Ahh, yeah. The `Collection` class implements PHP's [`Countable`](@if ($contestants->count())) interface but that (frustratingly!) doesn't apply for `empty()`... only for `count()`. Suggestion rescinded! :stuck_out_tongue_winking_eye:

app/Repositories/CacheUserRepository.php
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Get collection of users from cache by specified role or default to database lookup.
+     *
+     * @param  string|null $role
+     * @return \Illuminate\Support\Collection
+     */
+    public function getAllByRole($role = null)
+    {
+        $key = array_search($role, User::getRoles());
```
weerd: @DFurnes made the `$role` default to `null`, but the role of null is the key of `contestant` in regards to caching and when using gui things, so used this approach instead for grabbing the associated `key` for the `null` role ;)

DFurnes:
:+1:

weerd:
oh man it's happening?!

DFurnes:
:clap:

PR #75, opened 2016-03-11
Adds contest API 

### LINKED ISSUE #63: Create endpoint for Phoenix to get Contests
This endpoint can replace #26 

Phoenix should be able to hit this endpoint to see if there is a contest on a node and the response should include if the signup period on that contest is open (based on the dates on the waiting room).

deadlybutter:
#### What's this PR do?
- Added new endpoint to get a contest based on run nid
- Added new transformer for the unique response returned by ^
- Added new function to waiting room model to determine if it should be open for signups or not

Hit the endpoint http://homestead.app/api/v1/contests?run_nid=<run_nid>
Make sure the response looks good, something like this

screen shot 2016-03-11 at 11 31 37 am

Play around with changing the waiting room dates as well to make sure the signups_open works as intended

See conversation in #Laravel for why I made a new transformer 

Fixes #63


app/Http/Controllers/Api/ContestsController.php
```diff
@@ -19,13 +25,37 @@ class ContestsController extends ApiController
...
+     * @param int $run_nid
+     * @return \Illuminate\Http\Response
+     */
+    public function getByRunId($run_nid)
+    {
+        // Get contest by run nid
+        $contest = Contest::where('campaign_run_id', $run_nid)->firstOrFail();
+
+        // Get waiting room for this contest
+        $waitingRoom = WaitingRoom::where('contest_id', $contest->id)->firstOrFail();
```
DFurnes: I think this could be done with a relation on the model, like `$contest->waitingRoom`.

app/Http/Controllers/Api/ContestsController.php
```diff
@@ -19,13 +25,37 @@ class ContestsController extends ApiController
     public function __construct()
     {
         $this->transformer = new ContestTransformer;
+        $this->contestAndWaitingTransformer = new ContestAndWaitingTransformer;
```
weerd: So I think you'd just pass the contest data into the ContestTransformer and waiting room data into the WaitinRoomTransformer instead of creating a new transformer that repeats dealing with that data... @DFurnes mind chiming in?

deadlybutter: I can have two transformers in one response?

app/Http/Controllers/Api/ContestsController.php
deadlybutter: well thats just magic :sparkles:

app/Http/Controllers/Api/ContestsController.php
```diff
@@ -19,13 +25,37 @@ class ContestsController extends ApiController
...
+        // Get waiting room for this contest
+        $waitingRoom = WaitingRoom::where('contest_id', $contest->id)->firstOrFail();
+
+        // Construct a response with all of the data
+        $data = [
+            'waitingRoom' => $waitingRoom,
+            'contest' => $contest,
+        ];
+
+        return $this->item($data, 200, [], $this->contestAndWaitingTransformer);
```
DFurnes: This seems like it should be a nested object on the `ContestTransformer`... [check out the docs](http://fractal.thephpleague.com/transformers/) for how that'd work in Fractal (with the waiting room fields, like `signup_start_date` and `signup_end_date` living on the `WaitingRoomTransformer`).

weerd:
can you summarize the convo in #Laravel in the description so we don't need to go finding it if looking back at this issue?

weerd:
Thoughts on having 1 `signup` property and the `open`, `start` and `end` be nested properties inside it?

app/Models/WaitingRoom.php
```diff
@@ -29,6 +29,19 @@ public function users()
...
     }
 
+    /**
+     * Determines if the waiting room is open for signupsOpen
+     *
+     * @return bool whether the room is open or not
+     */
+    public function isOpen()
+    {
+        $start = Carbon::parse($this->signup_start_date)->startOfDay();
```
DFurnes: These can both be casted to Carbon objects by the model, so you can do `$this->signup_start_date->startOfDay()`.

app/Http/Transformers/ContestAndWaitingTransformer.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    public function transform($data)
+    {
+        $waitingRoom = $data['waitingRoom'];
+        $contest = $data['contest'];
+
+        return [
+            'signups_open' => $waitingRoom->isOpen(),
+            'signup_start' => $waitingRoom->signup_start_date,
+            'signup_end' => $waitingRoom->signup_end_date,
```
weerd: converto to iso using
```
$waitingRoom->signup_start_date->toIso8601String()
```

app/Models/WaitingRoom.php
```diff
@@ -29,6 +29,19 @@ public function users()
...
+        $end = Carbon::parse($this->signup_end_date)->endOfDay();
+        $today = Carbon::now();
+        return ($today >= $start) && ($today <= $end);
```
DFurnes: This might be more readable if you use Carbon's comparison methods. `$today->between($start, $end)`.

deadlybutter:
Yeah I asked in #Laravel if i could include custom vars in the response but didn't really get a clear answer. I guess I'm still not fluent in the terminology.

Also @weerd its the latest messages in laravel but for future reference

screen shot 2016-03-11 at 11 42 06 am

app/Models/WaitingRoom.php
weerd: They're already casted so that should work :)

```
// Models/WaitingRooms.php
/**
     * The attributes that should be mutated to dates.
     *
     * @var array
     */
    protected $dates = ['signup_start_date', 'signup_end_date'];
```

app/Http/Controllers/Api/ContestsController.php
DFurnes: Check out the docs I linked in my comment below for nested resources. :)

weerd:
Why do we need `/run` in the URL for the endpoint?

app/Models/WaitingRoom.php
deadlybutter: I did that and they didn't work if today == end or today == start, which according to Fantini in #competitions should be the intended functionality

deadlybutter:
@weerd /contests/<run_nid> didnt make sense to me, i figured any number after /contests should be the contest ID

app/Models/WaitingRoom.php
DFurnes: Hmm, are you sure? The `between` method should default to `>=` and `<=` for comparison.

weerd:
@deadlybutter so based on description in issue:

> Phoenix should be able to hit this endpoint to see if there is a contest on a node and the response should include if the signup period on that contest is open (based on the dates on the waiting room).

My initial thought would be the endpoint would be something like this:
```
// would normally return 1 item
/api/v1/contests?campaign_run_id=:id
```

Eventually there might be:
```
// could return a collection
/api/v1/contests?campaign_id:id
```

But they'd all be part of the main index request?

Not a fan of adding extra url segments in general unless absolutely necessary so we follow standard RESTful routing approaches, but `/run` sounds confusing to me cause it could mean the contest run or campaign run?

thoughts?

deadlybutter:
oh if URL params is the norm I guess we could do that

deadlybutter:
Ok removed the custom transformer and nested the stuffs into the contests. Looks a lot better now i think (thanks @weerd !)

cc: @DFurnes

deadlybutter:
Going to make it a URL param like Diego suggested and should be good for re-review

deadlybutter:
@weerd @DFurnes alrighty think its good for round 2

app/Http/Transformers/ContestTransformer.php
```diff
@@ -17,7 +17,11 @@ public function transform(Contest $contest)
     {
         return [
             'id' => $contest->id,
-            // @TODO: More to come...
+            'campaign_id' => $contest->campaign_id,
+            'campaign_run_id' => $contest->campaign_run_id,
+            'signups_open' => $contest->waitingRoom->isOpen(),
```
DFurnes: Not necessary for merge, but this could be placed on a `WaitingRoomTransformer` and [included as a nested resource](http://fractal.thephpleague.com/transformers/#including-data) so you could re-use the waiting room transformer elsewhere as well.

weerd:
@joe update your description w/ new url stuffs now that it's changed :dancer:

app/Models/WaitingRoom.php
```diff
@@ -29,6 +29,20 @@ public function users()
...
+     *
+     * @return bool whether the room is open or not
+     */
+    public function isOpen()
+    {
+        $start = $this->signup_start_date->startOfDay();
+        $end = $this->signup_end_date->endOfDay();
+        $today = Carbon::now();
+
+        return ($today >= $start) && ($today <= $end);
```
DFurnes: Did you try `return $today->between($start, $end)`? It should work for this and reads a lot nicer.

app/Http/Controllers/Api/ContestsController.php
```diff
@@ -22,10 +25,17 @@ public function __construct()
     }
 
     /**
-     * Get a collection of waiting rooms.
+     * Get a collection of contests
      */
-    public function index()
+    public function index(Request $request)
     {
+        $run_nid = $request->query('run_nid');
```
weerd: If you're only using the `$run_nid` in one place, no need for saving it as a variable.

(also can we maintain the naming we use elsewhere and just call it `$campaign_run_id`)

app/Http/Transformers/ContestTransformer.php
deadlybutter: true, but i dont think we will never need to transform this elsewhere

app/Http/Transformers/ContestTransformer.php
```diff
@@ -17,7 +17,11 @@ public function transform(Contest $contest)
...
+            'signup_start' => $contest->waitingRoom->signup_start_date->toIso8601String(),
+            'signup_end' => $contest->waitingRoom->signup_end_date->toIso8601String(),
```
weerd: Before when we had a WaitingRoomTransformer, i had it like this, which is more of what I meant to keep a response similar to Phoenix too:

```
        return [
            'id' => $waitingRoom->id,
            'campaign' => [
                'id' => $waitingRoom->campaign_id,
                'campaign_run' => [
                    'id' => $waitingRoom->campaign_run_id,
                ],
            ],
            'signup_dates' => [
                'start' => $waitingRoom->signup_start_date->toISO8601String(),
                'end' => $waitingRoom->signup_end_date->toISO8601String(),
            ],
            'created_at' => $waitingRoom->created_at->toISO8601String(),
            'updated_at' => $waitingRoom->updated_at->toISO8601String(),
        ];
```

app/Http/Controllers/Api/ContestsController.php
deadlybutter: there is two lines under it using the variable? would you prefer i just did request->query('run_nid') twice?

weerd:
Defs looks a lot better :)

app/Http/Transformers/ContestTransformer.php
deadlybutter: oh i can format it like that

deadlybutter:
new output per @weerd suggestions

```
{
  "data": {
    "id": 1,
    "campaign": {
      "id": 206,
      "campaign_run": {
        "id": 1687
      }
    },
    "waiting_room": {
      "signups_open": true,
      "signup_start": "2016-03-10T00:00:00+0000",
      "signup_end": "2016-03-11T00:00:00+0000"
    }
  }
}
```

also addressed the other comments, should be good to go

deadlybutter:
@weerd @DFurnes is this :+1: ?

weerd:
Awesome-sauce :dancer: :+1:

PR #81, opened 2016-03-14
Adds export to CSV option

### LINKED ISSUE #70: Pull user info into a CSV
since we are holding off with messaging, need a way to get user info to mail merge send reminders and leaderboard updates
refs #23

deadlybutter:
#### What's this PR do?
Allows user to export a 
- CSV of all Waiting room users
- all competitions in a contest 
- all users in a competition 

click the export button & verify the data matches the DB

Fixes #70

### DIFF
--- a/app/Models/WaitingRoom.php
+++ b/app/Models/WaitingRoom.php
@@ -29,7 +29,13 @@ class WaitingRoom extends Model
         return $this->belongsToMany(User::class);
     }
 
-    public function getCSVExport() {
+    /**
+     * Gets a CSV export of all the users in this waiting room.
+     *
+     * @return \League\Csv\ $csv
+     */
+    public function getCSVExport()
+    {
         $csv = \League\Csv\Writer::createFromFileObject(new \SplTempFileObject());
 
         $csv->insertOne(['id']);
--- a/app/Models/WaitingRoom.php
+++ b/app/Models/WaitingRoom.php
@@ -29,6 +29,18 @@ class WaitingRoom extends Model
         return $this->belongsToMany(User::class);
     }
 
+    public function getCSVExport() {
+        $csv = \League\Csv\Writer::createFromFileObject(new \SplTempFileObject());
+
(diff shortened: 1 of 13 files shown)


deadlybutter:
note: currently this only exports the ID's of users, once the cache work is done I can go back and add user details

deadlybutter:
also debating if i should include a time stamp in the filename, so if they export multiple times theyll know the latest version

app/Models/WaitingRoom.php
```diff
@@ -30,6 +30,24 @@ public function users()
     }
 
     /**
+     * Gets a CSV export of all the users in this waiting room.
+     *
+     * @return \League\Csv\ $csv
+     */
+    public function getCSVExport()
```
sbsmith86: These three `getCSVExport()` functions seem repetitive to me. Is there a way to make this a helper function in `helpers.php` so anything can access it?

deadlybutter: I agree it could be a helper, my only worry is im not sure down the road how long this feature will be used and what it will be exporting. EG: will the competitions export also show reportback progress? cause thats something which wont be inside the waiting rooms

sbsmith86: hmmm could the helper just take in an array of the data that the you want to export? That way it can be flexible, and the controller would just be responsible for creating that array of data.

deadlybutter: OOoooo thats a good idea. :+1:

deadlybutter:
@sbsmith86 added a helper function & updated the model exports to use it

app/helpers.php
```diff
@@ -28,3 +28,14 @@ function matchEmailDomain($email, $domain = 'dosomething.org')
 
     return false;
 }
+
+function buildCSV($data)
```
sbsmith86: Comments?

deadlybutter: :facepalm:

sbsmith86:
:100:

PR #84, opened 2016-03-14
Updates contest form

### LINKED ISSUE #77: update the contest create form
it would be a lot easier to create a waiting room at the same time as a contest, so let's add the waiting room start and end dates to the contest form

deadlybutter:
#### What's this PR do?
Updates the contest form to handle signup start / end
Updates controller logic to support ^

When you create a new competition / update one, it should also change the date on the waiting room

Need to do what Dave suggested here real quick to make the inputs pre-fill data https://dosomething.slack.com/archives/laravel/p1457979854000026

Fixes #77

### DIFF
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -37,7 +37,10 @@ class ContestsController extends Controller
      */
     public function store(ContestRequest $request)
     {
-        $contest = Contest::create($request->all());
+        $dateParams = ['signup_start_date', 'signup_end_date'];
+
+        $contest = Contest::create($request->except($dateParams));
+        $contest->WaitingRoom->fill($request->only($dateParams))->save();
 
         return redirect()->action('ContestsController@show', $contest->id)->with('status', 'Contest has been saved!');
     }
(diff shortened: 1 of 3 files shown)


sbsmith86:
So we decided to not use the `ContestServiceProvider` for this? I think this looks ok, but 1) we will need to remove that functionality. and 2) I still don't understand why this is the right way to do this. The good thing about attaching the waiting room creation to the `created()` eloquent event is that it ensures the contest has been created before making the waiting room that has a foreign key constraint on it.

deadlybutter:
In theory I could attach it to the service provider, but I wasn't sure how to get the extra request parameter in there

sbsmith86:
That's what I was sorta outlining [here](https://github.com/DoSomething/gladiator/issues/77#issuecomment-196334177) 

But i'm not gonna press it too much. If this works ok, then let's just go with this, just make sure to remove the service provider so that we don't create two waiting rooms.

deadlybutter:
this shouldn't be creating two waiting rooms though? its accessing the one made by the service provider if I'm understanding this right

deadlybutter:
Is this PR good to merge?

sbsmith86:
:+1:

PR #85, opened 2016-03-14
Fixes case in class reference

mshmsh5000:
#### What's this PR do?
Converted a class reference to upper case (`competition` to `Competition`).

This may address https://github.com/DoSomething/phoenix/issues/6170#issuecomment-196513424

Strange 500 errors when POSTing to the `/api/v1/users` method. The specific error is 

```
FatalErrorException in [compiled.php line 10379]():Class 'Gladiator\Models\competition' not found
```

Fixes DoSomething/phoenix/issues/6170

### DIFF
--- a/app/Models/Contest.php
+++ b/app/Models/Contest.php
@@ -26,7 +26,7 @@ class Contest extends Model
      */
     public function competitions()
     {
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

PR #90, opened 2016-03-15
Fix a bug keeping user from signing up for multiple competitions.

### LINKED ISSUE #88: Signing up for multiple contests with the same user causes app to fail
```
Illuminate\Database\QueryException">QueryException</abbr> in "/var/www/gladiator/releases/20160315170527/vendor/laravel/framework/src/Illuminate/Database/Connection.php line 669" 
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry &#039;559442cfa59dbfc9578b4c14&#039; for key &#039;PRIMARY&#039; (SQL: insert into `users` (`id`, `role`, `updated_at`, `created_at`) values (559442cfa59dbfc9578b4c14, , 2016-03-15 17:14:30, 2016-03-15 17:14:30))
```

image

weerd:
#### What's this PR do?
This PR updates a bug that was keeping API POSTed users from signing up for multiple competitions.

Try posting the same user to various competitions on different contests via the API. Get crazy error? No? Then it works!

The `Registrar` class method was returning an basic object with the user info from Northstar, instead of an instance of the `User` model if they are already in the system. In the methods that call this `Registrar` method, we do a check to see if the returned info is an `instanceof User` and if it is, then doesn't try to add them again to the system but continues on its way. But by returning a standard object, the check was failing and thus assuming the user didn't already exists in Gladiator and trying to add them again :scream: 

Fixes #88

---

### DIFF
--- a/app/Services/Registrar.php
+++ b/app/Services/Registrar.php
@@ -62,7 +62,7 @@ class Registrar
      * Resolve user account within Northstar/Gladiator system.
      *
      * @param  array  $credentials
-     * @return \Gladiator\Models\User|string
+     * @return \Gladiator\Models\User|object
      * @throws \Gladiator\Services\Northstar\Exceptions\NorthstarUserNotFoundException
      */
     public function findUserAccount($credentials)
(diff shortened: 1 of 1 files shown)


deadlybutter:
looks good to me

PR #91, opened 2016-03-15
Adds getAll() method to UserRepositoryContract and implementors

### LINKED ISSUE #89: Expose a getAll() method in UserRepository
Need to be able to grab all users, or pass an array of Northstar user IDs and grab the info for those users.

weerd:
#### What's this PR do?
This PR exposes a `getAll()` on the `UserRepositoryContract` which allows you to call the method from the class that implements the interface, and either pass an array of ids or nothing and it will return a collection of users.

For a quick and dirty run, hijack the UserController `index()` action and try hitting `$this->repository->getAll();` and seeing what it returns. Should be a rather large collection of users! And then try it the same thing but pass an array of Northstar IDs.

Rush job! Need to unblock @sbsmith86 and @deadlybutter so the code isn't super pretty, but gets the job done. I need to go back in and refactor a bit, because while working on this issue, it exposed a flaw in my `resolveMissingUsers()` method that checks a cached collection to make sure all the users requested were found in the cache, otherwise it goes and finds them using a single user `find()` method. You can imagine that on a large collection with lots of users missing from the cache (which can happen and is totally fine, don't sweat it :sweat_smile:) doing single requests for each of those User's for their Northstar profile info is inefficient. Best to switch it up to do a batched request! So that'll be an upgrade :dancer: 

Fixes #89

---

### DIFF
--- a/app/Repositories/CacheUserRepository.php
+++ b/app/Repositories/CacheUserRepository.php
@@ -60,6 +60,42 @@ class CacheUserRepository implements UserRepositoryContract
     }
 
     /**
+     * Get collection of all users or set of users by ids from cache or
+     * default to alternate respository lookup.
+     *
+     * @param  array $ids Northstar IDs
+     * @return \Illuminate\Support\Collection
+     */
+    public function getAll(array $ids = [])
+    {
+        // @TODO: This is messy and needs another pass.
+        if ($ids) {
+            $users = $this->retrieveMany($ids);
+
+            if (! $users) {
+                $users = $this->repository->getAll($ids);
+
+                if ($users->count()) {
+                    $group = $users->keyBy('id')->all();
+                    $this->storeMany($group);
+                }
+            } else {
+                $users = $this->resolveMissingUsers($users);
+                $users = collect(array_values($users));
+            }
+
+            return $users;
+        }
+
+        // @TODO: find a better way to retrieve all from cache.
+        // Will resort to just passing off to alternate repository
+        // and grabbing everything from Northstar without caching.
+        $users = $this->repository->getAll();
+
+        return $users;
+    }
+
+    /**
      * Get collection of users from cache by specified role or default to
      * alternate repository lookup.
      *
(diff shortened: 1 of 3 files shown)


weerd:
Not sure why StyleCI is yelling at me :confused:

app/Repositories/CacheUserRepository.php
```diff
@@ -60,6 +60,42 @@ public function find($id)
...
+     * Get collection of all users or set of users by ids from cache or
+     * default to alternate respository lookup.
+     *
+     * @param  array $ids Northstar IDs
+     * @return \Illuminate\Support\Collection
+     */
+    public function getAll(array $ids = [])
+    {
+        // @TODO: This is messy and needs another pass.
+        if ($ids) {
```
deadlybutter: nit picky i guess, but shouldnt this be !empty($ids) ?

app/Repositories/CacheUserRepository.php
```diff
@@ -60,6 +60,42 @@ public function find($id)
...
+            $users = $this->retrieveMany($ids);
+
+            if (! $users) {
```
deadlybutter: do we normally put a space between ! and $var ?

app/Repositories/CacheUserRepository.php
weerd: Empty array evaluates to `false` ;)

app/Repositories/CacheUserRepository.php
weerd: Yep, part of the StyleCI rules!

deadlybutter: moved the convo to slack cause it shouldnt be blockin your PR from merging

app/Repositories/CacheUserRepository.php
```diff
@@ -197,11 +233,23 @@ protected function retrieve($key)
...
+        $data = Cache::many($keys);
+
+        foreach ($data as $item) {
+            if ($item) {
+                $retrieved[] = $item;
+            }
+        }
+
+        if (count($retrieved)) {
+            return $data;
```
deadlybutter: what happens when this doesnt return anything? I feel like the return should just be outside of IF regardless of the count, no harm in returning an empty array

weerd: I had it that way and StyleCI yelled at me, thus the convo in Slack ;)

weerd: @joe It actually does return `null` when the function doesn't return anything. But by being more explicit it would have avoided this whole convo which is why I was asking about it in our slack convo :wink:

deadlybutter: yeah i gotcha

sbsmith86:
This is fine for merge for me.

deadlybutter:
:+1:

PR #92, opened 2016-03-16
Show Users in Waiting Rooms and Competitions

### LINKED ISSUE #82: need view to see users in waiting room on the contest view
It'd be helpful to know how many users are in a waiting room (and ideally who they are) before deciding to split or not.

Ideally this could be a new column on https://qa-gladiator.dosomething.org/waitingrooms where the number of users displays in the table and links to a view of who those users are.

Number of people in the waiting room should be the priority. This way Hillary and Chris can know how much traction the competition has received and if they should split it sooner rather than later given demand.

sbsmith86:
#### What's this PR do?

This adds user tables to the waiting room view and the competition view that show the users that are signed up and the users that are in a competition. Also adds totals to the waiting rooms and competition index views to show and over view of how many people are in each one. 

NEXT STEPS: I think we need to add pagination to these. 

Addresses #57 (still need to show user activity for each user)
Fixes #82

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -48,7 +48,7 @@ class CompetitionsController extends Controller
     {
         $contest = Contest::find($competition->contest_id);
 
-        $users = $this->repository->getAll($competition->users->lists('id')->toArray());
+        $users = $this->repository->getAll($competition->users->pluck('id')->toArray());
 
         return view('competitions.show', compact('competition', 'contest', 'users'));
     }
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -5,14 +5,23 @@ namespace Gladiator\Http\Controllers;
 use Gladiator\Http\Requests\CompetitionRequest;
 use Gladiator\Models\Competition;
 use Gladiator\Models\Contest;
+use Gladiator\Repositories\UserRepositoryContract;
 
 class CompetitionsController extends Controller
 {
     /**
+     * UserRepository instance.
+     *
+     * @var \Gladiator\Repositories\UserRepositoryContract
+     */
+    protected $repository;
+
+    /**
      * Create new CompetitionsController instance.
      */
-    public function __construct()
+    public function __construct(UserRepositoryContract $repository)
     {
+        $this->repository = $repository;
         $this->middleware('auth');
         $this->middleware('role:admin,staff');
     }
@@ -39,8 +48,7 @@ class CompetitionsController extends Controller
     {
         $contest = Contest::find($competition->contest_id);
 
-        // Get competition users.
-        $users = $competition->users;
+        $users = $this->repository->getAll($competition->users->lists('id')->toArray());
 
         return view('competitions.show', compact('competition', 'contest', 'users'));
     }
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -39,7 +39,10 @@ class CompetitionsController extends Controller
     {
         $contest = Contest::find($competition->contest_id);
 
-        return view('competitions.show', compact('competition', 'contest'));
+        // Get competition users.
+        $users = $competition->users;
+
+        return view('competitions.show', compact('competition', 'contest', 'users'));
     }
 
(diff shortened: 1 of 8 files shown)


app/Models/User.php
```diff
@@ -120,36 +120,4 @@ public static function hasNorthstarAccount($type, $id)
...
-                'signup' => null,
-            ];
-        } else {
-            $info += [
-                'name'  => $user->user_id,
-            ];
-        }
-
-        return $info;
-    }
```
weerd: Yaais! I like removing more methods _out_ of the Models :smile:

app/Http/Controllers/CompetitionsController.php
```diff
@@ -39,7 +48,9 @@ public function show(Competition $competition)
     {
         $contest = Contest::find($competition->contest_id);
 
-        return view('competitions.show', compact('competition', 'contest'));
+        $users = $this->repository->getAll($competition->users->pluck('id')->toArray());
```
angaither: I bet @weerd is going to have some choice words about `toArray()` :wink:

sbsmith86: @angaither lol ok. I was just following @weerd's [lead!](https://github.com/DoSomething/gladiator/blob/035239d1ca206f76297ac41a6924e57f98a4e7d9/app/Repositories/DatabaseUserRepository.php#L92)

app/Models/User.php
angaither: where is the home now?

app/Http/Controllers/CompetitionsController.php
angaither: :scream: he used to be angry at me for that. maybe it's different when it's coming from the cache and I know nothing.

app/Models/User.php
sbsmith86: @angaither it was a dumb function that we don't need anymore now that we are just pulling users from cache

app/Http/Controllers/CompetitionsController.php
weerd: We actually do want an array here, so all is well @angaither! Alternatively could use the `->all()` method on the collection which just grabs the items property.

weerd: The Cache retrieval requires an actual array of ids and you can't pass a collection. I would get angry when you would change the collection to an array when passing it to the view, haha :flushed:

weerd: I might end up changing [this](https://github.com/DoSomething/gladiator/blob/035239d1ca206f76297ac41a6924e57f98a4e7d9/app/Repositories/DatabaseUserRepository.php#L92) to use `->all()` but not really a huge deal. Just less processing if expected `items` property is always contains an array anyhoo.

app/Repositories/CacheUserRepository.php
```diff
@@ -208,7 +208,7 @@ protected function resolveUpdatedRoles($id, $role)
     protected function resolveMissingUsers($users)
     {
         foreach ($users as $key => $value) {
-            if ($value === false) {
+            if ($value === false or $value === null) {
```
weerd: @sbsmith86 just noting here that I created an issue #93 to look into why this is weird and can give different expected results to require checking for both

resources/views/waitingrooms/show.blade.php
```diff
@@ -31,6 +31,30 @@
...
+                            <th class="table__cell">Email</th>
+                            <th class="table__cell">Phone</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        @foreach($users as $key => $user)
+                            @if ($user)
+                                <tr class="table__row">
+                                    <td class="table__cell"><a href="{{ route('users.show', $user->id) }}">{{ $user->first_name or $user->id }}</td>
+                                    <td class="table__cell">{{ $user->email or NULL }}</td>
```
weerd: Would it be better for the `or` to default to empty string? Not sure.

sbsmith86: ¯\_(ツ)_/¯ seems to work ok? Not sure which is better either

weerd: K. Not sure either. We can always update later :)

weerd:
:+1:

PR #102, opened 2016-03-16
Updates exports

### LINKED ISSUE #98: Update CSV export
- First Name
- Last Name
- Email
- Cell
- Reportback admin link
- Flagged Status
- Quantity        
- Northstar User ID

deadlybutter:
#### What's this PR do?
Updates the competition and waiting room functionality to show user details

Does it export the right things?

Even though its not DRY right now eventually the exports of these two models will differ (pending reportback data stuff)

Fixes #98

### DIFF
--- a/app/Models/Competition.php
+++ b/app/Models/Competition.php
@@ -27,25 +27,4 @@ class Competition extends Model
     {
         return $this->belongsToMany(User::class);
     }
-
-    /**
-     * Gets a CSV export of all the users in this competition.
-     *
-     * @param \Gladiator\Repositories\UserRepositoryContract $repository
-     * @return \League\Csv\ $csv
-     */
-    public function getCSVExport($repository)
-    {
-        $data = [];
-        $users = $this->users;
-
-        // TODO: Competition/Reportback activity
-        array_push($data, ['northstar_id', 'first_name', 'last_name', 'email', 'cell']);
-        foreach ($users as $user) {
-            $userInfo = $repository->find($user->id);
-            array_push($data, [$user->id, $userInfo->first_name, $userInfo->last_name, $userInfo->email, $userInfo->mobile]);
-        }
-
-        return buildCSV($data);
-    }
 }
--- a/app/Models/Competition.php
+++ b/app/Models/Competition.php
@@ -31,16 +31,19 @@ class Competition extends Model
     /**
      * Gets a CSV export of all the users in this competition.
      *
+     * @param \Gladiator\Repositories\UserRepositoryContract $repository
      * @return \League\Csv\ $csv
      */
-    public function getCSVExport()
+    public function getCSVExport($repository)
     {
         $data = [];
         $users = $this->users;
 
-        array_push($data, ['id']);
+        // TODO: Competition/Reportback activity
+        array_push($data, ['northstar_id', 'first_name', 'last_name', 'email', 'cell']);
         foreach ($users as $user) {
-            array_push($data, $user->id);
+            $userInfo = $repository->find($user->id);
+            array_push($data, [$user->id, $userInfo->first_name, $userInfo->last_name, $userInfo->email, $userInfo->mobile]);
         }
 
         return buildCSV($data);
--- a/app/Services/ContestManager.php
+++ b/app/Services/ContestManager.php
@@ -49,5 +49,4 @@ class ContestManager
 
         return buildCSV($data);
     }
-
 }
--- a/app/Services/ContestManager.php
+++ b/app/Services/ContestManager.php
@@ -23,13 +23,20 @@ class ContestManager
         $this->repository = $repository;
     }
 
-    public function exportCSV($model, $includeReportbackData = false)
+    /**
+     * Build CSV data for a given WaitingRoom or Competition
+     *
+     * @param WaitingRoom|Competition $model
+     * @param bool $reportbacks Should this csv include reportback data?
+     * @return \League\Csv $csv
+     */
+    public function exportCSV($model, $reportbacks = false)
     {
         $data = [];
         $users = $model->users;
 
         $headers = ['northstar_id', 'first_name', 'last_name', 'email', 'cell'];
-        if ($includeReportbackData) {
+        if ($reportbacks) {
             //TODO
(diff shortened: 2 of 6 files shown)


deadlybutter:
@weerd or @sbsmith86 did either of you wanna take a look at this?

app/Models/Competition.php
```diff
@@ -31,16 +31,19 @@ public function users()
     /**
      * Gets a CSV export of all the users in this competition.
      *
+     * @param \Gladiator\Repositories\UserRepositoryContract $repository
      * @return \League\Csv\ $csv
      */
-    public function getCSVExport()
+    public function getCSVExport($repository)
```
sbsmith86: feels odd to take in the `$repository` here. Shouldn't it just pass data? If you need the `$repository` shouldn't you instantiate it in the Model instead? @weerd thoughts on this?

deadlybutter: @weerd said :no_good: to instantiating the repo in the model, and the internets agreed with me. this was the quickest solution i could think of :/

sbsmith86: What about using the repository in the controller to get the data you need and then passing that into this function?

deadlybutter: well then i might as well move the entire thing into the controller, but i thought we didn't want to clutter the controllers up? i dunno what our best practices are for this stuff

sbsmith86: Ok well, If this seems right to everyone else then :+1:

weerd: @deadlybutter this is where I was also mentioning instead of doing this in the Model, maybe we create a `ContestService` which has a bunch of functionality for dealing with contests, competitions, waitingrooms that don't feel right in a Model (which usually deals with a single item) and to help abstract crazy logic in controllers that might end up getting repeated. Could be a universal service that deals with this stuff; doesn't need to be a separate service each for contests, competitions, waitingrooms...

DFurnes: Yeah, this feels like a service class kinda thing to me. :+1:

weerd: maybe @DFurnes has some input from his experience in Northstar & Aurora?

image

DFurnes: :speech_balloon: :white_check_mark:

weerd: Haha, you commented while I was searching for a pic of Northstar & Aurora :stuck_out_tongue_winking_eye:

deadlybutter: gotcha, i guess now that we're doing more OO stuff my Java days are coming back to me, and doing something like this was :ok: 

time to figure out how to make a service thing

deadlybutter: gotcha, i guess now that we're doing more OO stuff my Java days are coming back to me, and doing something like this was :ok: 

time to figure out how to make a service thing

weerd: We have a few services in repo already. Check the Registrar as an example that doesn't to be formally defined as a Laravel Service Provider, since it'll likely be a simple class with lots of methods.

deadlybutter:
@weerd @sbsmith86 @DFurnes hows this? added a service called `ContestManager`

app/Services/ContestManager.php
```diff
@@ -0,0 +1,52 @@
...
+     * Create new Registrar instance.
+     *
+     * @param Northstar $northstar
+     */
+    public function __construct(UserRepositoryContract $repository)
+    {
+        $this->repository = $repository;
+    }
+
+    public function exportCSV($model, $includeReportbackData = false)
```
weerd: Add some docblocks and could probs just name `$includeReportbackData = false` -> `$reportbacks = false`

app/Services/ContestManager.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Gladiator\Services;
+
+use Gladiator\Repositories\UserRepositoryContract;
+
+class ContestManager
```
weerd: @sbsmith86 was chatting with @deadlybutter IRL regarding the name for this service. Would just `Manager` be better and it could handle methods for contests, competitions, waitingrooms? or does `ContestManager` work for all those? What do you think?

weerd: @deadlybutter defs about to use this service after you merge! :dancer:

sbsmith86: hmm yea maybe a `Manager` service and then if we have more specific needs per Model we can break them out?

weerd: :+1:

deadlybutter:
@sbsmith86 @weerd renamed to manager!

app/Services/Manager.php
```diff
@@ -0,0 +1,59 @@
...
+     * Build CSV data for a given WaitingRoom or Competition
+     *
+     * @param WaitingRoom|Competition $model
+     * @param bool $reportbacks Should this csv include reportback data?
+     * @return \League\Csv $csv
+     */
+    public function exportCSV($model, $reportbacks = false)
+    {
+        $data = [];
+        $users = $model->users;
```
weerd: Nice. I like this :)

app/Services/Manager.php
```diff
@@ -0,0 +1,59 @@
...
+        $headers = ['northstar_id', 'first_name', 'last_name', 'email', 'cell'];
+        if ($reportbacks) {
+            //TODO
+            array_push($headers, 'some reportback headers go here');
+        }
+
+        array_push($data, $headers);
+
+        foreach ($users as $user) {
+            $userInfo = $this->repository->find($user->id);
```
weerd: I don't think there'd be an issue overwriting the `$user` var here so no need to add extra var name:

```
$user = $this->repository->find($user->id);
```

sbsmith86:
Look much better to me! :100: :tada:

app/Services/Manager.php
weerd: and then when grabbing info it's cleaner `$user->first_name`, etc

weerd:
Nice! Just a quick comment on cleaning up the `$user` var but it's looking good :dancer:

weerd:
Snazzy! Nice work :tada: :+1:

PR #113, opened 2016-03-18
log in not login

angaither:
#### What's this PR do?
Updated menu to say `Log In` instead of `login`

are there spaces?  :space_invader: 

We use Log In and Log Out on other sites.
Keep it consistent.


weerd:
good :eyes: :+1:

PR #115, opened 2016-03-18
Set default value of signup date to date, or null.

### LINKED ISSUE #114: Contest create fails
When viewing `contests/create` see the following error
```ErrorException in 1ff0402878f3e966aa6579326bd8aa37b02628a8.php line 8:
Undefined variable: contest (View: /home/vagrant/Code/gladiator/resources/views/contests/partials/_form_contest.blade.php) (View: /home/vagrant/Code/gladiator/resources/views/contests/partials/_form_contest.blade.php)```

### PR DESCRIPTION
**Pull request by angaither**
#### What's this PR do?
Fixes error on contests/create route

test creating and editing a contest. 
image

I tried putting `NULL` in but it broke the styles. 
image

Fixes #114

### DIFF
--- a/resources/views/contests/partials/_form_contest.blade.php
+++ b/resources/views/contests/partials/_form_contest.blade.php
@@ -4,12 +4,12 @@
 
 <div class="form-item -padded">
     <label class="field-label" for="signup_start_date">Signup Start Date:</label>
-    <input type="date" name="signup_start_date" id="signup_start_date" value={{ $contest->waitingRoom->signup_start_date or old($contest->waitingRoom->signup_start_date)}} class="text-field"></input>
+    <input type="date" name="signup_start_date" id="signup_start_date" value={{ $contest->waitingRoom->signup_start_date or 'MM/DD/YYYY' }} class="text-field"></input>
 </div>
 
 <div class="form-item -padded">
     <label class="field-label" for="signup_end_date">Signup End Date:</label>
-    <input type="date" name="signup_end_date" id="signup_end_date" value={{ $contest->waitingRoom->signup_end_date or old($contest->waitingRoom->signup_end_date)}} class="text-field"></input>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: resources/views/contests/partials/_form_contest.blade.php [hunk 1]
```diff
@@ -4,12 +4,12 @@
 
 <div class="form-item -padded">
     <label class="field-label" for="signup_start_date">Signup Start Date:</label>
-    <input type="date" name="signup_start_date" id="signup_start_date" value={{ $contest->waitingRoom->signup_start_date or old($contest->waitingRoom->signup_start_date)}} class="text-field"></input>
+    <input type="date" name="signup_start_date" id="signup_start_date" value={{ $contest->waitingRoom->signup_start_date or 'MM/DD/YYYY' }} class="text-field"></input>
```
weerd: can't we set this as the `placeholder` attribute in the input? the `old()` method is useful cause it repopulates the field if there was a validation error and sends user back to the form.

angaither: it errors out b/c `$contest` isn't defined on the create route

weerd: Oh, it was used incorrectly! You only pass the string name of the item, not the actual property that might not exist. You can do this:

```php
<input type="date" name="signup_start_date" id="signup_start_date" value={{ $contest->waitingRoom->signup_start_date or old('signup_start_date', 'MM/DD/YYYY')}} class="text-field"></input>
```

If an old value exists, then it'll be in sitting in a property in the request object which holds an array of `old` data, and you grab it by passing a string of what it's called.

weerd: I do a similar approach in the user create form... uses the same partial as the edit form :dancer:

angaither:
@weerd updated!

weerd:
:older_man: :older_woman: `old()` :+1:

PR #118, opened 2016-03-18
Fixing assigned roles to cached users which were not having a role assigned.

### LINKED ISSUE #117: Bug with cached User roles when doing batch collections
Role assignment doesn't get appended to the Northstar user object when running batch collections, so it's missing from retrieved User objects from Repositories.

weerd:
#### What's this PR do?
This PR makes an update to fix an issue with batch collection of users. When running batch collections the User's role in the database wasn't getting assigned to the User object, whereas when doing a single User find it does get appended. This corrects that.

View the "Users" index page, click on an admin user and make sure in their view they are indicated as "admin" and not "member". Same for staff users.

Found what is a potential bug/feature in Northstar. When querying the `/users` endpoint on Northstar with a specific list of users, if any of the users aren't found, there is no indication. Talked to @DFurnes IRL and we discussed possible passing an additional parameter that could signify that we _do want to know_ what users couldn't be retrieved by returning `false` or `null` for that user in the resulting array response.

Fixes #117

---

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -46,8 +46,6 @@ class UsersController extends Controller
         $staff = $this->repository->getAllByRole('staff');
         $contestants = $this->repository->getAllByRole();
 
-        // dd($admins);
-        // dd($contestants);
         return view('users.index', compact('admins', 'staff', 'contestants'));
     }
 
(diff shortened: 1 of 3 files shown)


angaither:
bug/feature, eh?

angaither:
looks okay, but cc @sbsmith86 for changes here

sbsmith86:
Looks good to me! 👍

weerd:
@angaither none of the changes affect what @sbsmith86 did with showing User info in Competitions if that's what you mean :dancer:

PR #120, opened 2016-03-18
Minor menu meliorism

angaither:
#### What's this PR do?
add some linkage all over
update the default phoenix uri to exclude `/api/` because maybe we don't want that. 
remove delete buttons

#### How should this be manually tested?
check that things link, api still works

#### What are the relevant tickets?
Fixes #116, fixes #96

PR #121, opened 2016-03-21
User activity

### LINKED ISSUE #57: Add more info to competition user view

### LINKED ISSUE #108: User CSV export enhancements
For competition exports we need the following:

- Reportback admin link
- Flagged Status
- Quantity

This is needed by Monday so Hillary can send the first leaderboard update.

refs: #98

sbsmith86:
#### What's this PR do?

Show's user activity in the  `users.show` view for each competition that a user is in, and adds user activity to the competition CSV. 

NOTE: #57 asked for this to be shown in the competition view, but we decided for this first step, before any leaderboard behavior, we would show it on the user view.

Pull down, view a user that is in a competition, and you should see there reportback information, if they have submitted a reportback.

Currently this looks like: 

screen shot 2016-03-21 at 4 18 38 am

But we can certainly update this to show more information and/or show a more detailed version of the reportback activity in another route. This just shows the info that was asked of in the issues. 

Fixes #57 and Fixes #108 

@weerd @angaither @deadlybutter


.env.example
```diff
@@ -26,4 +26,5 @@ NORTHSTAR_URI=http://northstar.dev:8000
 NORTHSTAR_API_KEY=abc4324
 
 PHOENIX_URI=https://dev.dosomething.org:8888
+PHOENIX_API_URI=https://staging.beta.dosomething.org/api
```
deadlybutter: hm, why did we add two different urls?

sbsmith86: I had some merge conflicts around these lines, so I might not have fixed them correctly. Is it better to use dev urls in the example file or staging?

deadlybutter: hmm, i think staging just so it works out of the box. if someone needs to go gladiator / phoenix  dev then they can prob change the URL

deadlybutter:
CSV stuff looks good :+1: I'll leave northstar stuffs up to @weerd

.env.example
weerd: Yeah, plus I think you might need to do some funky local setup stuff to have the Phoenix dev running alongside the Gladiator dev and communicating with each other. Would rather keep staging uri. But in recent merged PR @angaither remove the `api` portion of the uri and is added when constructing the path, so this will like lead to a bug if kept here.

app/Services/Manager.php
```diff
@@ -37,9 +47,9 @@ public function exportCSV($model, $reportbacks = false)
         $users = $model->users;
 
         $headers = ['northstar_id', 'first_name', 'last_name', 'email', 'cell'];
+
         if ($reportbacks) {
```
weerd: Is it meant to be a single Reportback with multiple items? If so, maybe let's not pluralize this variable?

.env.example
sbsmith86: @weerd I think that is why i had merge conflicts, but it should be ok because looking at master, the  `PHOENIX_URI` variable is not being used anywhere but the services config, which I update in this PR to be in new `PHOENIX_API_URI` so we should be good there. I will update everything to point to staging though.

app/Services/Manager.php
deadlybutter: its just an indicator that I created in a previous PR so we know whether the CSV should add reportback data or not. (so just a boolean, no reportback data in it)

sbsmith86: I can update to `hasReportback` I was concerned about this to when I needed the extra check that we got a reportback back, it's a little repetitive now. That might make it more readable.

app/Services/Manager.php
```diff
@@ -80,4 +97,62 @@ public function collectContestInfo($data)
...
+     * @return object $signups
+     */
+    public function getUserSignup($id, $campaign = null, $campaignRun = null)
+    {
+        $signups = $this->northstar->getUserSignups($id, $campaign);
+
+        // Only return the sign up record for the run that was specified.
+        if ($campaignRun) {
+            foreach ($signups as $key => $signup) {
+                if ($signup->campaign_run->id == $campaignRun) {
```
weerd: use `===`

.env.example
```diff
@@ -25,5 +25,6 @@ MAIL_ENCRYPTION=null
 NORTHSTAR_URI=http://northstar.dev:8000
 NORTHSTAR_API_KEY=abc4324
 
-PHOENIX_URI=https://dev.dosomething.org:8888
+PHOENIX_URI=https://staging.beta.dosomething.org
+PHOENIX_API_URI=https://staging.beta.dosomething.org/api
```
angaither: @sbsmith86 I updated this in my [last PR](https://github.com/DoSomething/gladiator/pull/120/files#diff-7fdf976827c0f65f17c003adf2ba56d9R16) so the api thing should work without adding a second var for this.

app/Services/Manager.php
sbsmith86: @weerd i was having some issues with this because the campaign run id that comes back from the northstar call is an integer, and the campaign run id that comes back from the gladiator db is a string.

.env.example
sbsmith86: oooo ok ok i see.

.env.example
angaither: I'd prefer if we actually use Dev url's here
DS-Homestead is set up to use phoenix correctly and there's nothing funky about using those at the same time

sbsmith86: lol :ok_hand: will update!

config/services.php
```diff
@@ -30,8 +30,8 @@
     ],
 
     'phoenix' => [
-        'uri' => env('PHOENIX_URI'),
-        'version' => 'v1',
+        'uri' => env('PHOENIX_API_URI'),
```
angaither: this should be changed back to `PHOENIX_URI`

app/Services/Manager.php
weerd: Hmmm, maybe cast one of them to a specific type and then check it?

```
if ((string) $signup->campaign_run->id === $campaignRun) { ... }
```

Would prefer to have a DS standard where we always use `===` cause it's more secure and highlights bugs :dancer:

app/Services/Manager.php
```diff
@@ -80,4 +97,62 @@ public function collectContestInfo($data)
...
+
+        $signup = $this->getUserSignup($id, $campaign, $campaign_run);
+
+        if ($signup && $signup->reportback) {
+            // Provide the admin URL to the reportback.
+            $signup->reportback->admin_url = env('PHOENIX_URI') . '/admin/reportback/' . $signup->reportback->id;
+
+            // Format the update timestamp
+            $signup->reportback->updated_at = new Carbon($signup->reportback->updated_at);
+            $signup->reportback->updated_at = $signup->reportback->updated_at->format('Y-m-d');
```
weerd: Could this be combined w/ the line above?

```
 $signup->reportback->updated_at = new Carbon($signup->reportback->updated_at)->format('Y-m-d');
```

Not actually sure if that works...

sbsmith86: it doesn't work, i tried. I guess the Carbon constructor doesn't allow you to call a function on what it returns.

app/Services/Manager.php
```diff
@@ -80,4 +97,62 @@ public function collectContestInfo($data)
...
+     * @return object $signups
+     */
+    public function getUserSignup($id, $campaign = null, $campaignRun = null)
+    {
+        $signups = $this->northstar->getUserSignups($id, $campaign);
+
+        // Only return the sign up record for the run that was specified.
+        if ($campaignRun) {
+            foreach ($signups as $key => $signup) {
+                if ($signup->campaign_run->id === (string) $campaignRun) {
```
weerd: Wait, isn't `$campaignRun` already coming in as a string?

```
@param  string $campaignRun Campaign Run ID
```

And it's the northstar campaign_run id that needs to be changed to a string? Or is it vice versa? If so, update the docblock too, since it's an `int` coming in.

app/Services/Manager.php
weerd: Cool. Good to know!

app/Services/Manager.php
sbsmith86: $campaignRun is coming in as an int sorry for the confusion. I will update the docs.

weerd: Ah, ok no worries :)

weerd:
After the docs update, should be alllllls good :+1:

PR #125, opened 2016-03-21
App now responds properly with appropriate error responses.

### LINKED ISSUE #119: Update errors responses from API
Exceptions are thrown during errors when hitting the API but not being caught and set as proper API errors.

weerd:
#### What's this PR do?
This PR updates responses so if an API related error occurs, the app now responds with an appropriate error response formatted in JSON, and if the same error occurs via the web app we can now differentiate and respond accordingly to that as well.

For the Northstar user not found error, we now have responses that deal with both avenues.

**API Error Response:**
image

**Web App Error Response:**
image

Also added a 404 page template which I tried at first for the web app but decided best to redirect back to the form with a status message instead, but doesn't hurt to keep it.

Try passing a an email you know won't exist on Northstar to create a new user via both the API POST request and via the add User form in the web app and see if the response matches what is indicated above.

Proper error responses! :dancer: 

_Note:_ followed a similar approach to what's going on in Northstar :)

Fixes #119

---

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -46,7 +46,6 @@ class Handler extends ExceptionHandler
      */
     public function render($request, Exception $e)
     {
-
         if ($request->ajax() || $request->wantsJson()) {
             return $this->json($request, $e);
         }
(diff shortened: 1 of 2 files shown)


DFurnes:
Lookin' good. :+1:

weerd:
Another _note_, just like @DFurnes setup in Northstar, the debug property in API response is only available if the app is in `debug` mode ;)

PR #127, opened 2016-03-22
default value for flagged status

### LINKED ISSUE #126: Set flagged as 'pending' as a default
When the `flagged` property of a reportback is returned as `false` from the Northstar API, set the value to `pending`

sbsmith86:
#### What's this PR do?

Sets the flagged status to `pending` when it is returned as false from the API.

screen shot 2016-03-21 at 8 08 30 pm

Fixes #126

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -148,6 +148,9 @@ class Manager
             $signup->reportback->updated_at = new Carbon($signup->reportback->updated_at);
             $signup->reportback->updated_at = $signup->reportback->updated_at->format('Y-m-d');
 
+            // Return set flagged status to 'pending' if it is false.
+            $signup->reportback->flagged = ($signup->reportback->flagged) ? $signup->reportback->flagged : 'pending';
+
             // Return the reportback.
             return $signup->reportback;
         }


app/Services/Manager.php
```diff
@@ -148,6 +148,9 @@ public function getUserActivity($id, $model)
             $signup->reportback->updated_at = new Carbon($signup->reportback->updated_at);
             $signup->reportback->updated_at = $signup->reportback->updated_at->format('Y-m-d');
 
+            // Return set flagged status to 'pending' if it is false.
+            $signup->reportback->flagged = ($signup->reportback->flagged) ? $signup->reportback->flagged : 'pending';
```
weerd: Just curious, why did you have to wrap it in parentheses? Pre-emptive strike against possible issue or was there a specific issue?

weerd:
:+1:

app/Services/Manager.php
sbsmith86: oh well i guess that's just how I always do ternary's, especially when im just testing if a variable isset, null, or false, since it reads more like a test to me. But if that's not standard I can remove them!

PR #129, opened 2016-03-22
Add key authorization for post requests to Gladiator.

### LINKED ISSUE #69: Add API Key for POST requests authorization
Forgot to add this in #48

weerd:
#### What's this PR do?
This PR updates the API to add a middleware for authorizing specific request methods. This update adds a conditional for `POST` requests that requires supplying a `X-DS-Gladiator-API-Key` in the header of the request with a specified key as the value. 

If the key matches what is on Gladiator then the request is accepted, otherwise it is rejected and responds with an error, which is caught by the `Handler` class and responds with a formatted JSON API error.

The exception thrown is a new `UnauthorizedAccessException` which could be utilized for other conditions in the web app as well regarding unauthorized access to data. 

Pull down branch, `composer dump-autoload`, make sure to add the new `GLADIATOR_API_KEY=uxm107oct1977` to your **.env** file and then try making a POST request to the api with a valid api key, and then with an invalid one. Does the API respond as expected and described above?

There's a new env variable that needs to be added to prod, qa, and local env for the API key:

```
GLADIATOR_API_KEY=uxm107oct1977
```

Fixes #69

---


weerd:
Forgot to update the API docs. Will do that now!

app/Http/Controllers/Api/UsersController.php
```diff
@@ -27,6 +27,19 @@ public function __construct(Registrar $registrar)
     {
         $this->registrar = $registrar;
         $this->transformer = new UserTransformer;
+
+        $this->middleware('auth.api');
```
deadlybutter: should all of the API controllers have this?

weerd: Probs. I can add to them as well. Was concentrating on the Users controller, but makes sense to add to all of them now.

DFurnes:
Looks good to me! :+1:

angaither:
@weerd we should wait on this till phoenix issue is done?

deadlybutter:
well i cant do the phoenix issue till this is merged.

we're not actively doing anything (competitions related) between the two platforms so shouldn't be a big deal to merge this now.

weerd:
@angaither yeah defs. Just wanted to push up the code so it could be checked. I'll hold off merging until I get the go ahead from you that Phoenix is ready :)

weerd:
If the issue w/ Competitions and the CRON on Phoenix is resolved and there's no active issues then should be ok to merge. What say you @angaither?

deadlybutter:
@weerd im assuming andrea is referring to the issue she just assigned me https://github.com/DoSomething/phoenix/issues/6300

weerd:
@deadlybutter @DFurnes @angaither I updated the `Api\ContestsController` to include the new middleware, and also added docs for `/contests` endpoint since they were missing.

I also did an update to change the name of the optional query param passed to the `/contests` endpoint since it was named after a very Drooooopaly name, and want to stay away from doing that because we don't do it anywhere else in Gladiator :dancer:

app/Http/Controllers/Api/ContestsController.php
```diff
@@ -22,16 +22,28 @@ class ContestsController extends ApiController
...
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
      */
     public function index(Request $request)
     {
-        $campaign_run_nid = $request->query('run_nid');
-        if (isset($campaign_run_nid)) {
-            $contest = Contest::with('waitingRoom')->where('campaign_run_id', $campaign_run_nid)->firstOrFail();
+        $campaignRunId = $request->query('campaign_run_id');
```
deadlybutter: :+1:

deadlybutter:
new naming looks good to me, code looks ok as well but might wanna get a +1 from andrea/shae/dave as well

.env.example
```diff
@@ -11,6 +11,8 @@ CACHE_DRIVER=redis
 SESSION_DRIVER=file
 QUEUE_DRIVER=sync
 
+GLADIATOR_API_KEY=uxm107oct1977
```
sbsmith86: will this key work on our local?

weerd:
cool, sounds good. thanks @deadlybutter

.env.example
weerd: Yeah, this key is for local dev etc. We'll be setting a more snazzy complex one for prod.

angaither:
:+1:

PR #133, opened 2016-03-22
Northstar exception update

weerd:
#### What's this PR do?
This PR removes the need to do `instanceof` checks for the API error responses. During PR #129 found it's easier to extend the `HttpException` class from Symphony and pass the error code in the custom exception class definition and then in the Handler, if the exception is of that type, we can just grab the error code with `$e->getStatusCode()` in a more generic fashion. Using a similar approach for other http exceptions will help eliminate the need for a slew of `if` conditions in the `Handler` class.

When creating a bogus user with email not on Northstar, or when using the API to create a user with email not on Northstar, do you get appropriate error responses? Great!

None, just helpful cleanup.

---

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -72,10 +72,6 @@ class Handler extends ExceptionHandler
     {
         $code = 500;
 
-        // if ($e instanceof NorthstarUserNotFoundException) {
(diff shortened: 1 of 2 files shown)


deadlybutter:
kool :shaved_ice:

PR #134, opened 2016-03-22
Update to not show all users in waiting room.

### LINKED ISSUE #131: Waiting room shows users even when empty
### background

When viewing a waiting room, after it has been split and the users removed, the waiting room show view is still showing users:

screen shot 2016-03-22 at 4 44 04 pm

The problem is with the call to get the users in the `show()` method in the `WaitingRoomsController`

```
$users = $this->repository->getAll($room->users->pluck('id')->toArray());
```

`$room->users` returns an empty collection and when the `getAll()` method doesn't recieve anything, it just returns all users. 

### Solution

The `getAll()` method needs to be updated accept a `null` param and not return anything.

weerd:
#### What's this PR do?
This PR updates the `WaitingRoomsController` `show()` method, so that it first checks to see if the search for users in the specific waiting room returns an empty array, then don't try and go to the repository to grab the users. The default for the `getAll()` method `$ids` param is an empty array, which signals that method to go ahead and instead of grabbing a bunch of users based on specified ids, to just grab alllll the users.

The `WaitingRoomsController` was sending through an empty array which was leading to the above condition. So if there's no one in the waiting room after a split action, the users is an empty array and we should just return that to the `view()` instead of even bothering with the repository.

Load up a waiting room that's already split (and not accepting new signups)... there should be no users in there.

I was contemplating adding a conditional that removes the **Signups** table from inside the Waiting Room show template, but decided that maybe it's best to leave it showing empty and Signups as `0` so it's more obvious that there are no signups:

screen shot 2016-03-22 at 6 13 58 pm

Fixes #131

---

### DIFF
--- a/app/Http/Controllers/WaitingRoomsController.php
+++ b/app/Http/Controllers/WaitingRoomsController.php
@@ -69,9 +69,15 @@ class WaitingRoomsController extends Controller
      */
     public function show(WaitingRoom $room)
     {
+        $users = [];
+
         $contest = Contest::find($room->contest_id);
 
-        $users = $this->repository->getAll($room->users->pluck('id')->toArray());
+        $ids = $room->users->pluck('id')->toArray();
+
+        if ($ids) {
+            $users = $this->repository->getAll($ids);
+        }
 
         return view('waitingrooms.show', compact('room', 'contest', 'users'));
     }


app/Http/Controllers/WaitingRoomsController.php
```diff
@@ -69,9 +69,15 @@ public function store(WaitingRoomRequest $request)
      */
     public function show(WaitingRoom $room)
     {
+        $users = [];
+
         $contest = Contest::find($room->contest_id);
 
-        $users = $this->repository->getAll($room->users->pluck('id')->toArray());
+        $ids = $room->users->pluck('id')->toArray();
```
deadlybutter: this pluck this, laravel why are you so great

weerd: image

sbsmith86:
:100:

PR #135, opened 2016-03-23
Messages schema and model.

### LINKED ISSUE #130: Create a messaging table
As an admin, I want to be able to send messages to people in a contest

potential db structure 

id | contest id | type | subject | body
| --- | --- | --- | --- | ---|
auto int | int | enum | varchar | long text 
1 | 5 | reminder | something | something

weerd:
#### What's this PR do?
This PR adds the migration for upcoming Messages work. It also adds a `Message` model and a helper function to extract the different message types when needed, and adds associated relationship method to the `Contest` model.

Just :eyes: the code. Nothing to test just yet! **Most importantly**, closely review the migration and make sure that all looks correct... should there be an `index()` on the `type`? I figured it'd be useful, but also not a MySQL :neckbeard: 

We were initially thinking of using `enum` for the `type` column data type, but it could end up complicating things, especially when adding/removing different email types in the future and needing to do that all with costly migrations :scream_cat: 

Fixes #130

---

### DIFF
--- a/app/Models/Message.php
+++ b/app/Models/Message.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Gladiator\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class Message extends Model
+{
+    /**
+     * Array of available message types.
+     *
+     * @var array
+     */
+    protected static $types = [
+        'reminder',
+    ];
+
+    /**
+     * Get the list of Message types.
+     *
+     * @return array
+     */
+    public static function getTypes()
+    {
+        return static::$types;
+    }
+}
--- a/app/Models/Message.php
+++ b/app/Models/Message.php
@@ -12,8 +12,8 @@ class Message extends Model
      * @var array
      */
     protected static $types = [
-        'leaderboard_update',
         'leaderboard_final',
+        'leaderboard_update',
         'reminder',
         'reminder_first',
         'reminder_last',
--- a/app/Models/Message.php
+++ b/app/Models/Message.php
@@ -12,8 +12,21 @@ class Message extends Model
      * @var array
      */
     protected static $types = [
+        'leaderboard_update',
+        'leaderboard_final',
         'reminder',
+        'reminder_first',
+        'reminder_last',
+        'welcome',
     ];
+
+    /**
+     * Get the contest associated with this message.
+     */
+    public function contest()
+    {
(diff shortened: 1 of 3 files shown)


app/Models/Message.php
```diff
@@ -0,0 +1,40 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+class Message extends Model
+{
+    /**
+     * Array of available message types.
+     *
+     * @var array
+     */
+    protected static $types = [
```
angaither: I dunno if we should have so many types @ first

database/migrations/2016_03_23_175405_create_messages_table.php
```diff
@@ -0,0 +1,35 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('messages', function (Blueprint $table) {
+            $table->increments('id');
+            $table->integer('contest_id')->unsigned()->index();
+            $table->foreign('contest_id')->references('id')->on('contests')->onDelete('cascade');
+            $table->string('type')->index();
+            $table->subject('subject')->nullable();
```
angaither: `  $table->subject('subject')->nullable();` subject != a database type. `text` or `string`

app/Models/Message.php
weerd: Cool. Yeah, nice thing with doing this as an array on the model it's easy to update. I was going off of the **Gladiator - Messaging** list of template Sr. Harbur set up.

deadlybutter: is there a harm in having them?

angaither: yeah, I'm fine with the array on the model, but I think we should just start with `reminder` for now. it seems unlikely that will be different from other `reminder` emails

weerd: Yeah true, so I'll do generic types and the body copy can always just differ but there are defined general _types_. Good call @angaither

database/migrations/2016_03_23_175405_create_messages_table.php
weerd: oh gawd! :facepalm:

weerd: Maybe _should_ be a database type! haha :crying_cat_face:

angaither:
:ship:

PR #136, opened 2016-03-23
Remove duration, set end date on split

### LINKED ISSUE #94: Remove duration  and set end dates on the competition
this should no longer be metadata attached to the contest, the duration or end dates should be set when you make the competition

> As an admin I want to select the end date after I split the waiting room

sbsmith86:
#### What's this PR do?
* Removed the duration field on contests and all references to the duration field. 
* Add field to split form that admins can use to set the end date for competitions when they do a split. Also added all the validation, split logic updates to accommodate this. 
* Updated the split form view a bit. It didn't seem useful to list all of the users in each competition at this point in the flow, but I can either add this back in later if folks really need it, or we can maybe link out to another view that has a list of all the users. 

The new split view looks like this:

screen shot 2016-03-23 at 2 43 08 pm

serena-split

Pull down this branch 
Run `composer update` to get the `doctrine/dbal` package
Run `php artisan:migrate` to do the db updates
Check that duration field is gone-zo.

Then go to split a waiting room, 
you should be able to set a date that is after today as the end date for the competitions.

NOTE: The `doctrine/dbal` package is [required in order to do column drops](https://laravel.com/docs/5.2/migrations#dropping-columns).
 
674702c838b

Fixes #94 

### DIFF
--- a/app/Http/Controllers/WaitingRoomsController.php
+++ b/app/Http/Controllers/WaitingRoomsController.php
@@ -127,7 +127,7 @@ class WaitingRoomsController extends Controller
         $contest = Contest::find($room->contest_id);
         $room->saveSplit($contest, $split);
 
-        return redirect()->route('waitingrooms.index')->with('status', 'Waiting Room has been split!');
+        return redirect()->route('contests.show', $room->contest_id)->with('status', 'Waiting Room has been split!');
     }
 
     /**
--- a/app/Http/Controllers/WaitingRoomsController.php
+++ b/app/Http/Controllers/WaitingRoomsController.php
@@ -111,6 +111,11 @@ class WaitingRoomsController extends Controller
     {
         $split = $room->getDefaultSplit();
 
+        // Get user info from Northstar.
+        foreach ($split as $key => $users) {
+            $split[$key] = $this->repository->getAll($users);
+        }
+
         return view('waitingrooms.split', compact('split', 'room'));
     }
 
--- a/app/Http/Controllers/WaitingRoomsController.php
+++ b/app/Http/Controllers/WaitingRoomsController.php
@@ -7,6 +7,7 @@ use Gladiator\Models\WaitingRoom;
 use Gladiator\Http\Requests\WaitingRoomRequest;
 use Gladiator\Repositories\UserRepositoryContract;
 use Gladiator\Services\Manager;
+use Gladiator\Http\Requests\SplitRequest;
 
 class WaitingRoomsController extends Controller
 {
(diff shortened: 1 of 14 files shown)


deadlybutter:
i :heart: the new split room look. old one was basically my debuging interface

angaither:
notable PR. 
extreme success.
:ship:

weerd:
@sbsmith86 totally not necessary, but you could use a ternary like on the Contest show page to decide whether to pluralize or not (1 Proposed Competition \ 2 Proposed Competitions) :)

https://github.com/DoSomething/gladiator/blob/master/resources/views/contests/show.blade.php#L40

app/Http/routes.php
```diff
@@ -40,10 +40,10 @@
     ]);
 
     // Waiting rooms routes.
+    Route::model('waitingrooms', 'Gladiator\Models\WaitingRoom');
```
deadlybutter: I feel like these we're intentionally below the custom routes, do they still work?

app/Http/Controllers/WaitingRoomsController.php
```diff
@@ -121,13 +122,13 @@ public function showSplitForm(WaitingRoom $room)
      * @param  \Gladiator\Models\WaitingRoom  $room
      * @return \Illuminate\Http\Response
      */
-    public function split(WaitingRoom $room)
+    public function split(SplitRequest $request, WaitingRoom $room)
     {
         $split = $room->getDefaultSplit();
         $contest = Contest::find($room->contest_id);
-        $room->saveSplit($contest, $split);
+        $room->saveSplit($contest, $split, $request->competition_end_date);
```
deadlybutter: should we add an ->endOfDay() to this?

app/Http/routes.php
weerd: The `Route::model()` shouldn't matter (I don't think). The custom routes def needed to be before the `Route::resource('waitingrooms'..)`, but I think that's it.

app/Http/Controllers/WaitingRoomsController.php
sbsmith86: whoops good catch!

weerd:
Thoughts on removing the :banana: in the title :trollface:

deadlybutter:
:no_good:

app/Http/routes.php
sbsmith86: this was just a little OCD kicking in, it doesn't seem to make a difference functionally, but like it seemed out of place stuck in the middle of everything. All it is is defining the model to use with routes names 'waitingrooms' so you can do the type-hinting thing. Seemed better to get that definition out of the way first, and then deal with the route you want.

deadlybutter: :+1: as long as the routes remain in tact i dont have a problem with it :D

weerd:
Looks go to me after the `endOfDay()` update! :+1:

weerd:
also :+1: on the commit gif capture :wink:

deadlybutter:
:shipit:

PR #140, opened 2016-03-24
Update to fix bug with missing method if no password provided.

### LINKED ISSUE #139: Bug when logging in with no password
Gladiator throws an exception because it tries to call a non-existent method on the `Validator` class when trying to login without a password.

weerd:
#### What's this PR do?
This PR adds a simple fix to a bug when logging in without a password provided. The app throws an exception because when it handles the validation error, before returning to the page it tries to access the validation messages using an incorrect method.

Try logging in with an email and no password. Do you see crazy error-town or a proper validation error on the form indicating that a `password is required`?

Fixes #139

---

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -51,7 +51,7 @@ class Handler extends ExceptionHandler
         }
 
         if ($e instanceof ValidationException) {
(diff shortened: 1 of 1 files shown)


sbsmith86:
:100:

PR #141, opened 2016-03-24
update deploy steps, again 😈

angaither:
#### What's this PR do?
attempt numero dos on running migrations

noy idea

I think that's the pattern here, I have to say after X , then migrate.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -51,4 +51,4 @@ end
 
 after "deploy:update", "deploy:cleanup"
 after "deploy:symlink", "deploy:link_folders"
-after "deploy:artisan_migrate", "deploy:artisan_cache_clear"
+after "deploy:link_folders", "deploy:artisan_migrate"


config/deploy.rb
```diff
@@ -51,4 +51,4 @@
 
 after "deploy:update", "deploy:cleanup"
 after "deploy:symlink", "deploy:link_folders"
-after "deploy:artisan_migrate", "deploy:artisan_cache_clear"
+after "deploy:link_folders", "deploy:artisan_migrate"
```
weerd: do you need this `"deploy:artisan_cache_clear"`?

angaither: I think we added that in the voting app, to fix something, but thinking about it here, would just mean we would wipe user info on every deploy, so we probably don't want that.

DFurnes:
:+1: :gem:

config/deploy.rb
weerd: Ah, gotcha. Ok. Didn't know it was a vestige from the Voting App!

weerd:
:+1:

config/deploy.rb
DFurnes: Yeah, it’d be nice to have a “clean” starter Wercker & Cap config files. We still have some strange public folder linking step that I can't seem to remove that tries to do something to a `public/images` and `public/stylesheets` folder on every Northstar deploy. :grimacing:

angaither: yeah, happen in gladiator as well & I didn't see _where_ that was happening

DFurnes: Yeah, it's weird. Made an issue at DoSomething/devops#80 to track.

Issue #137, opened 2016-03-24
Allow admins to customize and create message content


angaither:
As an admin I would like to set up the text for messaging when I create a contest. 

__feelings__ 
- setting these should be optional, not required  :open_file_folder: 
- i would imagine this should also be ~smashable~ :octopus: 
- this should also handle basic text replacement of :repeat: 
  - first name :name_badge: 
  - end date :calendar:

weerd:
@angaither So not sure about the "smashable" since we're using Forge. The smashable was a JS component of Bootstrap. We could maybe add a new component for Forge though...

sbsmith86:
@weerd i think we should add it for forge-admin :hammer:

weerd:
Gah! Closing my issue @sbsmith86 :confounded: But yeah, good idea for making it a component for Forge-admin :dancer:

PR #144, opened 2016-03-25
Mail queue

### LINKED ISSUE #132: Implement laravel mail queue
add a mandrill send email functionality

(we can test/write this with some fake copy so it isn't blocked by #130)

sbsmith86:
#### What's this PR do?

* Added the `QueueMessageRequest` event and `QueueMessage` event listener that fire and listen when we need to send and email. The listener implements the `ShouldQueue` Interface that pushes the job onto the redis queue. 

* Adds a `MessagesController` that kicks of the event to send the message when called. 

* A very basic reminder email view for testing...so we get some content in there

* Updates to `.env.example` that show the new settings that we need to run the queue and mandril driver. 

Pull down branch. 
Update `.env` with correct settings. (mandril key is in mandrill, login with lastpass)
Set up a message in your DB

Run queue listener:
`php artisan queue:listen`

Run command in `art tinker`
`Gladiator\Http\Controllers\MessagesController::sendMessage($message_id)`

You should get an email...or actually, I, will get an email cause my email address is hardcoded right now.

This just does the basic set up work of sending emails in Gladiator, we will still have to add logic to send to actual people and need to wait for @weerd 's work to be finised on #137 

I also mentioned on slack that we are gonna have to boot up the [queue listener on gladiator deploys](https://github.com/DoSomething/gladiator/blob/master/config/deploy.rb#L46-L48) similar to what is done in [longshot](https://github.com/DoSomething/longshot/blob/5a20eb9383d2f11896bfd48998a4c8f3e61ac4f9/config/deploy.rb#L55)

Also noting that we should do some logging for succesful email sends and failures.

Fixes #132 


weerd:
> You should get an email...or actually, I, will get an email cause my email address is hardcoded right now.

:smile:

config/mail.php
```diff
@@ -15,7 +15,7 @@
     |
     */
 
-    'driver' => env('MAIL_DRIVER', 'smtp'),
+    'driver' => env('MAIL_DRIVER', 'log'),
```
angaither: do we wanna change these here or in our `.env` file?

app/Events/QueueMessageRequest.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Gladiator\Events;
+
+use Gladiator\Models\Message;
+use Illuminate\Queue\SerializesModels;
+
+class QueueMessageRequest extends Event
+{
+    use SerializesModels;
+
+    public $message;
```
weerd: docblock ;)

config/mail.php
sbsmith86: @angaither it is in my .env, i just had the log as the default for development. I can reset this though, we should probably default to mandrill.

app/Events/QueueMessageRequest.php
sbsmith86: ack

sbsmith86:
@angaither @weerd made some updates. Can you take a look again?

app/Listeners/QueueMessage.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param  QueueMessageRequest  $event
+     * @return void
+     */
+    public function handle(QueueMessageRequest $event)
+    {
+        $content = $event->message;
+
+        $type = $content->type;
+
+        Mail::send('messages.' . $type, ['content' => $content], function ($msg) use ($content) {
```
weerd: You probs don't need the `Mail` facade if you have it included in the constructor. Can you do:

```
$this->mail->send(...)
```

app/Listeners/QueueMessage.php
```diff
@@ -0,0 +1,45 @@
...
+
+            // @TODO - Pull from address from contest setting.
+            $msg->from('beyonce@app.com', 'Beyonce');
```
weerd: :joy:

app/Listeners/QueueMessage.php
sbsmith86: oh ya i forgot to update that when I moved this function to the listener.

config/mail.php
```diff
@@ -15,7 +15,7 @@
     |
     */
 
-    'driver' => env('MAIL_DRIVER', 'smtp'),
+    'driver' => env('MAIL_DRIVER', 'mandrill'),
```
weerd: You don't need to change these. The `smtp` string is what it defaults to in case you don't have the key in the `.env` :smiley:

config/queue.php
```diff
@@ -16,7 +16,7 @@
     |
     */
 
-    'default' => env('QUEUE_DRIVER', 'sync'),
+    'default' => env('QUEUE_DRIVER', 'redis'),
```
weerd: Same as above. No need to change this.

resources/views/messages/reminder.blade.php
```diff
@@ -0,0 +1 @@
+<?php echo $content->body ?>
```
weerd: Can't use blade? Or is this just placeholder for now?

If not, then probs don't need the `.blade` in the file name...

config/mail.php
sbsmith86: I know! but just felt like we should default to what we are expecting to use...i mean i doesn't matter i guess cause these things should always be set

resources/views/messages/reminder.blade.php
sbsmith86: @weerd This should definitely be blade, this is just a test to get things through. We eventually need more views, and the correct formatting for everything.

config/mail.php
weerd: yep. I read something, cause I had changed it when I did the redis caching stuff too, in the app config, and saw it's a bad idea to change _there_ from the Laravel defaults. Cause if redis is not turned on it can at least default to a less ideal but still working driver.

resources/views/messages/reminder.blade.php
weerd: Gotcha! thanks for clarifying!

config/queue.php
```diff
@@ -16,7 +16,7 @@
     |
     */
 
-    'default' => env('QUEUE_DRIVER', 'sync'),
+    'default' => env('QUEUE_DRIVER', 'database'),
```
weerd: Uno mas; should be `sync`

sbsmith86: @weerd updated!

weerd:
:+1: looks good

PR #149, opened 2016-03-28
Messaging in contest creation

### LINKED ISSUE #137: Allow admins to customize and create message content

weerd:
#### What's this PR do?
This PR allows creating the messages for a Contest when creating a new contest. When creating a new Contest, the message fields are pre-populated with some default content to simplify message creation.

Try creating a contest along with some messages. Make sure that all the messages are saved to the database with the correct `contest_id`.

This currently only allows creating the messages when the contest is created, but not editing them. That it coming in an upcoming PR.

Fixes #137

---

### DIFF
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -49,6 +49,10 @@ class ContestsController extends Controller
      */
     public function store(ContestRequest $request)
     {
+        dd($request->all());
+
+        // (new MessagesController)->store();
+
         $dateParams = ['signup_start_date', 'signup_end_date'];
 
         $contest = Contest::create($request->except($dateParams));
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -5,6 +5,7 @@ namespace Gladiator\Http\Controllers;
 use Gladiator\Models\Contest;
 use Gladiator\Services\Manager;
 use Gladiator\Http\Requests\ContestRequest;
+use Gladiator\Repositories\MessageRepository;
 
 class ContestsController extends Controller
 {
@@ -49,14 +50,15 @@ class ContestsController extends Controller
      */
     public function store(ContestRequest $request)
     {
-        dd($request->all());
+        $contest = Contest::create([
+            'campaign_id' => $request->input('campaign_id'),
+            'campaign_run_id' => $request->input('campaign_run_id'),
+        ]);
 
-        // (new MessagesController)->store();
+        $contest->waitingRoom->fill($request->only(['signup_start_date', 'signup_end_date']))->save();
 
-        $dateParams = ['signup_start_date', 'signup_end_date'];
-
-        $contest = Contest::create($request->except($dateParams));
-        $contest->WaitingRoom->fill($request->only($dateParams))->save();
+        $repository = new MessageRepository;
+        $repository->createMessagesForContest($contest, $request->input('messages'));
 
         return redirect()->action('ContestsController@show', $contest->id)->with('status', 'Contest has been saved!');
     }
--- a/app/Repositories/MessageRepository.php
+++ b/app/Repositories/MessageRepository.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Gladiator\Repositories;
+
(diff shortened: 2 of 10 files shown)


app/Repositories/MessageRepository.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param  array $data
+     * @return \Gladiator\Models\Message
+     */
+    public function create($contest, $data)
+    {
+        $message = new Message;
+        $message->type = $data['type'];
+        $message->key = $data['key'];
+        $message->subject = $data['subject'];
+        $message->body = $data['body'];
```
angaither: if these are all fillable, can't we `$message->data->save()` or whatever the laravel magic is? :crystal_ball:

resources/views/contests/create.blade.php
```diff
@@ -11,7 +11,34 @@
...
+                    {{ csrf_field() }}
+
+                    <h2 class="heading -alpha">Settings</h2>
+
                     @include('contests.partials._form_contest')
+
+
+                    <h2 class="heading -alpha">Messaging Settings</h2>
+
+                    @include('contests.partials._form_contest_messaging', ['type' => 'welcome', 'key' => '1', 'label' => 'Welcome'])
```
angaither: `@TODO: allow the type to be a dropdown & key is automatically assigned`

app/Repositories/MessageRepository.php
weerd: yeah defs! Needed to get it working first cause I had a different architecture before which was grabbing things from separate places and not all coming in as array :) still cleaning a few things up!

app/Repositories/MessageRepository.php
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @param  \Gladiator\Models\Contest $contest
+     * @param  array $data
+     * @return \Gladiator\Models\Message
+     */
+    public function create($contest, $data)
+    {
+        $message = new Message($data);
+
+        return $contest->messages()->save($message);
```
weerd: @angaither cleaner! 

Although I actually tried to do something like what we're doing in the `ContestController@store()` method, where we do:

```php
...
$contest->waitingRoom->fill($request->only(['signup_start_date', 'signup_end_date']))->save();
```

So I tried doing:

```php
$contest->messages->fill($data)->save();
```

But I kept getting an error that the `fill()` method didn't exist. Not sure why, but my suspicion is because the relationship is a `hasMany` and thus continues on with an expected collection of messages, as opposed to the `waitingRoom` example where that is a `hasOne`... but otherwise not sure. Do you or maybe @DFurnes know why it didn't think the method existed in this scenario?

angaither:
@weerd should we change it to refs issue # instead of fixes b/c of 
> This currently only allows creating the messages when the contest is created, but not editing them. That it coming in an upcoming PR.

weerd:
Figured we'd just create another issue to deal w/ editing, since #137 indicates mostly just the creation part.

> As an admin I would like to set up the text for messaging when I create a contest.

I was going to address in this PR, but I take it I'm potentially blocking other work by not getting the creation part in. So I can just do the read/edit in upcoming issue :)

app/Http/Controllers/ContestsController.php
```diff
@@ -49,10 +50,15 @@ public function create()
      */
     public function store(ContestRequest $request)
     {
-        $dateParams = ['signup_start_date', 'signup_end_date'];
```
deadlybutter: you didnt like my laravel parameter magic?

angaither:
:+1:

app/Http/Controllers/ContestsController.php
weerd: Well, the form now includes the new messaging inputs as well, so it was no longer `except()ing` just date param. Was pretty cool though, but with the changes, no longer made sense :wink:

PR #154, opened 2016-03-29
Add contest sender field

### LINKED ISSUE #143: Add send from value
As an admin I want to set who is sending the email, per contest. 

this includes: 
Name & email address

weerd:
#### What's this PR do?
This PR adds a "sender" field to the Contest creation form. It allows providing the Contest an email address that can later be used as the sender for the mail messages to contestants. It is _not_ required on initial Contest creation, and allows you to add a sender when you edit the Contest.

Try creating a contest and saving an email to the **Messages Sender** field! Then edit that contest and make sure the value you entered shows up in the field too :dancer: 

Fixes #143

---

### DIFF
--- a/app/Models/Contest.php
+++ b/app/Models/Contest.php
@@ -11,7 +11,7 @@ class Contest extends Model
      *
      * @var array
      */
-    protected $fillable = ['campaign_id', 'campaign_run_id'];
+    protected $fillable = ['campaign_id', 'campaign_run_id', 'sender'];
 
     /**
      * Get the waiting room associated with this contest.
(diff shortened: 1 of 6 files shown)


app/Models/Contest.php
```diff
@@ -54,4 +54,14 @@ public function getCSVExport()
...
     }
+
+    /**
+     * Set the sender attribute for the contest.
+     *
+     * @param string $value
+     */
+    public function setSenderAttribute($value)
+    {
+        $this->attributes['sender'] = ! empty($value) ? $value : null;
```
sbsmith86: we aren't storing this in the DB?

weerd: huh?

weerd: This is just a mutator so if the field isn't filled out it sets the item in the DB to `null` instead of an empty string.

sbsmith86: oh my bad. was confused cause i didn't see a migration for that field. was that in a different PR?

weerd: No worries! Yep. That happened in a prior PR along w/ the migrations for the Messages :dancer:

sbsmith86:
:+1:

PR #156, opened 2016-03-30
Remove users from competition

### LINKED ISSUE #151: Add remove user functionality
As an admin, I want to be able to remove a user from a competition so they no longer receive messaging. 

I'm not sure if we should actually `delete` the user or just remove them from the competition

angaither:
#### What's this PR do?
Add functionality on the `user.show` page to remove a user from a competition

Pull down, and go to a user show page, hit remove. Is that pivot table record removed?

This only removes the record in the pivot table, it doesn't remove the user from the database.
There's also no confirmation message before hitting this, that's why it's hidden all the way in the user show page. 

image
image

Fixes #151


sbsmith86:
:100:

PR #157, opened 2016-03-30
Leaderboad email day

### LINKED ISSUE #145: Create a leaderboard day variable
As an admin, when splitting a `contest` I want to specify which **day** a user will get `leaderboard` email updates, this is per `competition` 

Notes: 
- This isn't to automatically send the emails, it's to add it in as a variable in the content)
- This should be an actual day of the week. (sunday, monday, wednesday)
- Another variable we want in the `reminder` email is the day of the week, minus one
 - for example, leaderboards are sent on `Monday` and we want a reminder email to say, upload your rb by `Sunday`
 - I'm not even sure if we can do math on days of the week :elephant: :calendar:

sbsmith86:
#### What's this PR do?
Adds a `leaderboard_msg_day` field to `competitions` and allows admins to set this field when splitting a waiting room.

Go to split a waiting room and you should see a new dropdown that allows you to set the day you want the leaderboard to be sent. You should be able to submit this form with no issues. 

Fixes #145

### DIFF
--- a/app/Models/WaitingRoom.php
+++ b/app/Models/WaitingRoom.php
@@ -88,11 +88,15 @@ class WaitingRoom extends Model
 
     /*
      * Creates competitions based on the given user split.
+     *
+     * @param \Gladiator\Models\Contest $contest
+     * @param Array                     $split
+     * @param \Illuminate\Http\Request  $request
      */
-    public function saveSplit($contest, $split, $endDate)
+    public function saveSplit($contest, $split, $request)
     {
         $startDate = Carbon::now()->startOfDay();
-        $endDate = new Carbon($endDate);
+        $endDate = new Carbon($request->competition_end_date);
 
         foreach ($split as $competitionGroup) {
             // For each split, create a competition.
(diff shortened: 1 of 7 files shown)


angaither:
:sheep: :ship: :shipit:

app/Models/WaitingRoom.php
```diff
@@ -88,11 +88,15 @@ public function getDefaultSplit()
...
+     * @param \Gladiator\Models\Contest $contest
+     * @param Array                     $split
+     * @param \Illuminate\Http\Request  $request
      */
-    public function saveSplit($contest, $split, $endDate)
+    public function saveSplit($contest, $split, $request)
     {
         $startDate = Carbon::now()->startOfDay();
-        $endDate = new Carbon($endDate);
+        $endDate = new Carbon($request->competition_end_date);
```
weerd: Like this better, passing the full `$request` :dancer:

weerd:
¡OLÉ!
:sheep: it!

app/Models/WaitingRoom.php
DFurnes: It might make sense to use the Request's array-access interface here `$request['competition_end_date']` so that you could use an array for this parameter if you ever wanted to initiate this from somewhere other than a web request.

DFurnes: :japanese_ogre: He says post-merge! Just a thought for future refactoring!

weerd: @DFurnes if so, couldn't you just cast what you pass to be a simple object? unless that's like a bad practice or something :confused:

DFurnes: You could, but it seems like unnecessarily complicating things imo.

Issue #158, opened 2016-03-31
Add a field for sender name


angaither:
refs #143 

currently we only store the email address, not the name as well.

We _could_ create a new field and store them in the same col in the database, or create a new field for this.

Issue #152, opened 2016-03-29
Actually send the right messages


angaither:
Connect work from #144 & #149
allow admins to push a button in a competition to send the message they pick to the people in the competition

PR #159, opened 2016-03-31
Send those messages for fake

### LINKED ISSUE #155: Create a 'send test' email
As an admin, before I hit send for sending emails to all users in a comp, I want to send the email to myself and make sure it look :ok:

angaither:
#### What's this PR do?
- created a view to display all messages within a contest (on the competition page)
- created a `send` button to send the message to the person who created the contest
- made some super quick view templates that we should fancy up 

Pull down, send some emails

Fixes #155 
Refs #152 
needs work from #158

### DIFF
--- a/app/Events/QueueMessageRequest.php
+++ b/app/Events/QueueMessageRequest.php
@@ -15,6 +15,7 @@ class QueueMessageRequest extends Event
      * @var \Gladiator\Models\Message
      */
     public $message;
+    public $sender;
 
     /**
      * Create a new event instance.
@@ -24,9 +25,10 @@ class QueueMessageRequest extends Event
      *
      * @return void
      */
-    public function __construct(Message $message)
+    public function __construct(Message $message, $sender)
     {
         $this->message = $message;
+        $this->sender = $sender;
     }
 
     /**
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -7,6 +7,20 @@ use Gladiator\Events\QueueMessageRequest;
 
 class MessagesController extends Controller
 {
+
+
+    /**
+     * Display the specified resource.
+     *
+     * @param  string  $id  Contest ID
+     * @return \Illuminate\Http\Response
+     */
+    public function show($id)
+    {
+        $messages = Message::where('contest_id', '=', $id)->get();
+
+        return view('messages.show', compact('messages'));
+    }
     /**
      * Fire an event that queues a message to be sent.
      *
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -31,5 +31,7 @@ class MessagesController extends Controller
         $msg = Message::find($id);
 
         event(new QueueMessageRequest($msg));
+
+        return redirect()->route('messages.show', 1)->with('status', 'Fired that right the hell off!');
     }
 }
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -8,8 +8,6 @@ use Gladiator\Events\QueueMessageRequest;
 
 class MessagesController extends Controller
 {
-
-
     /**
      * Display the specified resource.
      *
@@ -22,6 +20,7 @@ class MessagesController extends Controller
 
         return view('messages.show', compact('messages'));
     }
+
     /**
      * Fire an event that queues a message to be sent.
      *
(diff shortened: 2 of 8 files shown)


app/Events/QueueMessageRequest.php
```diff
@@ -24,9 +25,10 @@ class QueueMessageRequest extends Event
      *
      * @return void
      */
-    public function __construct(Message $message)
+    public function __construct(Message $message, $sender)
```
sbsmith86: Now that I think about it, and not necessarily needed for merge, I wonder if we should just pass in an array or object that has all the info that we need about a message in it i.e the message, the sender, the list to send to. This way we only have to  grab one thing in the listener, and send the one array to the `Mail::send()` method, that can handle all the message configuration.

app/Http/routes.php
```diff
@@ -33,6 +33,9 @@
     Route::get('contests/{contest}/export', 'ContestsController@export')->name('contests.export');
     Route::resource('contests', 'ContestsController');
 
+    Route::resource('messages', 'MessagesController');
+    Route::get('messages/{message}/send', 'MessagesController@sendMessage')->name('messages.send');
```
sbsmith86: :fireworks:

app/Http/Controllers/MessagesController.php
```diff
@@ -2,20 +2,37 @@
...
+     *
+     * @param  string  $id  Contest ID
+     * @return \Illuminate\Http\Response
+     */
+    public function show($id)
+    {
+        $messages = Message::where('contest_id', '=', $id)->get();
+
+        return view('messages.show', compact('messages'));
+    }
```
weerd: Won't conflict, but I currently am working on a view for showing all the messages and editing them within a single collected view after clicking a button on the Contests page to "View & Edit Messages". Just noting here that we may want to consolidate.

app/Http/Controllers/MessagesController.php
```diff
@@ -2,20 +2,37 @@
...
+
+    /**
      * Fire an event that queues a message to be sent.
      *
      * @param int $id
      */
-    public static function sendMessage($id)
+    public static function sendMessage(Message $message)
```
weerd: How does Laravel know that the `Message` model should be bound if there's no definition in the routes to do the whole `Route::model('messages'...` thing?

sbsmith86: There were some issues with that but @angaither updated the route to be 

` Route::get('messages/{message}/send', 'MessagesController@sendMessage')->name('messages.send');`

and some laravel magic made it work. It seems that passing the "message" in to the route definition works like some sorta type-hinting .

angaither: so _i think_ it's because in the route I say I'm passing a message in the route `messages/{message}/send`

app/Http/Controllers/MessagesController.php
angaither: yeah!

app/Events/QueueMessageRequest.php
angaither: ooooo yeah that's a better route

angaither: I was going to work on who to send after this so i can fix that up in a different pr

sbsmith86:
Looks good to me :100:

app/Http/Controllers/MessagesController.php
weerd: Oh interesting. I was trying something similar and it wouldn't do that for me :( It's otherwise something that can be specified in the `RouteServiceProvider` I believe...

weerd:
:+1:

app/Http/Controllers/MessagesController.php
DFurnes: @weerd Laravel 5.2 does [automatic route model binding](https://laravel.com/docs/5.2/routing#route-model-binding) so you don't have to manually bind with `Route::model`, which is really cool. _...but_ for some reason it only injects the pluralized form of the resource when automatically binding `Route::resource` routes (so like it'd only work if you injected `Message $messages` into the other REST routes).

Luckily you can [opt-in to using non-plural variable names](https://github.com/DoSomething/nightwing/blob/1149a428b2e661ea3b98bdb3681f1e56005960f8/app/Providers/RouteServiceProvider.php#L27) in your provider but it's still weeeeird.

angaither: @DFurnes comin in with the :otwell: wisdom :crystal_ball:

PR #161, opened 2016-03-31
Updated system for how messages are handled in forms.

### LINKED ISSUE #150: Allow viewing & editing Messages post creating them
Refs #137

Creating the messages was completed in #149 but now need to be able to view them when viewing the contest (likely a link to it's own page of messages so it doesn't clutter the Contest view) and editing them.

weerd:
#### What's this PR do?
This PR adds the capability for admins to view the series of messages associated with a specific Contest and edit the messages. It also significantly cleans up the default messages approach. There is now a `Correspondence` class that helps deal with default messages and with messages in forms. This class is called via a helped method `correspondence()`, and does a few cool things like:

- collect all default messages
- build and output attributes for the form elements (like `id`, `name`, and `for` so they don't have to be manually created each time and eliminate typos
- outputs the correct message per field for each elements value, and first checks if there is any _old_ input, for example if an admin edited a message and submitted the create contest form but a validation error sends them back to the form... their edits to the message will be retained and repopulated in the field!

There is also a new view for seeing/editing all the messages for a contest at once which can be accessed via a "VIEW & EDIT MESSAGES" button in the Contest show page.

Try creating a contest, editing the values in the messages and causing a validation error so you are sent back to the form; did the edits to the messages stay?

Try successfully creating a contest and its messages, and then view the contest and edit the messages. Did the changes get saved?

Fixes #150

---
@DoSomething/gladiator


DFurnes:
:+1: from me!

app/helpers.php
```diff
@@ -1,6 +1,34 @@
 <?php
 
 /**
+ * Build a CSV from the supplied data.
+ *
+ * @param  array  $data
+ * @return \League\Csv $csv
+ */
+function buildCSV($data)
```
DFurnes: Nitpick: I think it's customary to have helper methods use `snake_case` in Laravel. Or could this be renamed to `csv` perhaps (kinda like the built-in `view()` or `json()` helpers)?

angaither:
:ship:

app/helpers.php
```diff
@@ -1,6 +1,34 @@
...
+    foreach ($data as $row) {
+        $csv->insertOne($row);
+    }
+
+    return $csv;
+}
+
+function correspondence($message = null, $field = null)
+{
+    $correspondence = app('Gladiator\Http\Utilities\Correspondence');
```
DFurnes: Oh, other silly style nitpick: it'd be nice to use `::class` here for static analysis goodness.

app/helpers.php
DFurnes: I also just realized that this was just moved up a few lines, so not at all necessary to change in this PR. But maybe worth a future little refactor! :)

weerd: Ah, good call. I wasn't actually adding this in my PR, just shifting it around to list function alphabetically. I'll create an issue, cause there's a few functions in there that need to be renamed to follow the standard!

app/helpers.php
weerd: oh jes! I forget it's a thing now.

Issue #153, opened 2016-03-29
Create variable/token standards


angaither:
We need a way to grab & replace the variables we are using

- standard way of naming things
  -  `[Model:attribute]` so an example `[Contest:end_date]` or just `:column_name`
  - also need to think about what we will call the things from phoenix
- an easy way to replace the content 

Tokens
---------
#### Gladiator
- `:sender_name`
- `:sender_email` : maybe not a token, but needed for the `from` var
- `:leaderboard_msg_day`
- `:leaderboard_msg_day-1`
- `:end_date` (competition end date) 
- `:pro_tip`

#### User cache 
- `:first_name` 

#### Phoenix cache
- `:campaign_title` 
- `:reportback_noun` 
- `:reportback_verb`
- `:link_to_campaign#prove` (this should be a better name but I wanted to ~describe~ it)

#### Calculated tokens
- `:top_three_reportbacks` (by quantity)
  - includes [`:first_name`, `:caption`, `.reportback_noun`, `:reportback_quantity`, `:reportback_photo`]
- `:leaderboard`

angaither:
we could also do [this](https://laravel.com/docs/4.2/templates#extending-blade)

angaither:
thinking about accessors as well, but I don't think we will have the variables we need to pass them in correctly
also when displaying the content, it'll appear to be replaced already. hmm

sbsmith86:
Don't know if this is helpful, but could we define an array of tokens/variables and then create a function that does the replacement? like this...

```
$replace = [':drink' => 'coffee', ':name' => 'Ricardo'];
$str = 'Hi my name is :name. I like to drink :drink.';

echo str_replace_dynamic($replace, $str);

function str_replace_dynamic(array $replace, $string) { 
   return str_replace(array_keys($replace), array_values($replace), $string);    
}
```

Found it here: http://laravel.io/forum/07-16-2014-how-do-i-replace-the-words-of-a-text

angaither:
@sbsmith86 yeah, that's a similar approach to what we did in the [scholarship app](https://github.com/DoSomething/longshot/blob/c4ae29de785a879abc9feb9dfdbf6a0cab25593d/app/models/Email.php#L15)

DFurnes:
Yeah, I like the solution @sbsmith86 posted. :+1: You could maybe use a `$tokenizable` property on each model class with the fields you want to be able to be accessed from tokens like this (or just allow anything and exclude `$hidden`)?

[...]

PR #164, opened 2016-03-31
Some tokens and better routes

angaither:
#### What's this PR do?
- Creates `tokenizable` items on the competition
- Find and replace those tokens in messages
- Change the route you send messages from to include info about the competition
- Sets standard to look for `:token` all lowercase and all matching the name that stored in the db
  - this means some default message copy will need to be updated

- pull down, add some messages with `:token`s and watch them replace like magic before being delivered to your inbox.

I only tested this with the few competitions tokens, not first names, not any phoenix data, figured we could get a proof of concept for this and go from there. 

Refs #153  Refs #152

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -7,6 +7,7 @@ use Gladiator\Models\Competition;
 use Gladiator\Models\Contest;
 use Gladiator\Repositories\UserRepositoryContract;
 use Gladiator\Services\Manager;
+use Gladiator\Models\Message;
 use Gladiator\Models\User;
 
 class CompetitionsController extends Controller
@@ -114,5 +115,19 @@ class CompetitionsController extends Controller
         $user->competitions()->detach($competition->id);
 
         return redirect()->back()->with('status', 'User was removed from competition ' . $competition->id);
+    }
+
+        /**
+     * Detach a user from a competition.
+     *
+     * @param  \Gladiator\Models\Competition  $competition
+     * @param  \Gladiator\Models\Message  $message
+     * @return \Illuminate\Http\Response
+     */
+    public function message(Competition $competition, Contest $contest)
+    {
+        $messages = Message::where('contest_id', '=', $contest->id)->get();
+
+        return view('messages.show', compact('messages'));
     }
 }
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -117,7 +117,7 @@ class CompetitionsController extends Controller
         return redirect()->back()->with('status', 'User was removed from competition ' . $competition->id);
     }
 
-        /**
+    /**
      * Detach a user from a competition.
      *
      * @param  \Gladiator\Models\Competition  $competition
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/CompetitionsController.php
```diff
@@ -115,4 +116,18 @@ public function removeUser(Competition $competition, User $user)
 
         return redirect()->back()->with('status', 'User was removed from competition ' . $competition->id);
     }
+
+    /**
+     * Detach a user from a competition.
```
sbsmith86: What is this function for? Is it just showing all messages? should this be moved into the `MessageController`?

app/Http/Controllers/MessagesController.php
```diff
@@ -29,10 +30,14 @@ public function show($id)
     public static function sendMessage(Message $message)
     {
         $contest_id = request('contest_id');
+        $competition_id = request('competition_id');
+
         $from = Contest::find($contest_id)->sender;
+        $competition = Competition::find($competition_id);
 
-        event(new QueueMessageRequest($message, $from));
+        //@TODO: only pass one thing in here instead of 3 or more.
```
sbsmith86: :+1:

app/Models/Message.php
```diff
@@ -37,4 +37,16 @@ public static function getTypes()
     {
         return static::$types;
     }
+
+    public static function prepareMessage($message, $model)
+    {
+        //@TODO: I only tested this with competition, is there other things we want to replace?
+        $tokens = $model::$tokenizable;
+        foreach ($tokens as $token) {
+            $message->body = str_replace(':'.$token, $model->$token, $message->body);
```
sbsmith86: Could we abstract this into a function that just takes an array of tokens and a string of text, so as we need to add more models and tokens we can just combine them all into one array and then use this new function to do the string replacement?

resources/views/messages/show.blade.php
```diff
@@ -26,8 +26,8 @@
                             <tr class="table__row">
                                 <td class="table__cell">{{ $message->label or 'N/A'}}</td>
                                 <td class="table__cell">{{ $message->subject or 'N/A'}}</td>
-                                <td class="table__cell">{{ str_limit($message->body, 100)}}</td>
-                                <td class="table__cell"><a href="{{ route('messages.send', ['message' => $message->id, 'contest_id' => $message->contest_id]) }}" class="button -tertiary ">Send</a></td>
+                                <td class="table__cell">{{ str_limit($message->body, 100) }}</td>
```
sbsmith86: should we link to an individual view of a message in case the admin wants to view the full message?

sbsmith86:
I like the approach! made some comments but if this is just a prototype then it looks good to me.

app/Http/Controllers/CompetitionsController.php
angaither: it's showing and sending messages in a competition. I originally had in in the message controller, but needed more info from the competition, happy to move it back if that makes more sense though

resources/views/messages/show.blade.php
angaither: probz, yeah

app/Models/Message.php
angaither: yeah I was going to play around with that when adding more types of replacement

app/Http/Controllers/CompetitionsController.php
weerd: This doesn't actually detach a user from a competition though right? Docblock copied over from above item?

PR #167, opened 2016-04-01
Update sender and bug fixes

### LINKED ISSUE #158: Add a field for sender name

weerd:
#### What's this PR do?
This PR updates the database schema for the Contest model to change `sender` to `sender_email` and also include a `sender_name`. It updates the associated Contest forms to include adding values for those fields as well as updating them.

It also updates a complication with casting dates from the database to Carbon instances. It seems that all Carbon instances include the time as well as the date, and when passing this to the HTML5 `<input type="date" />` element. We only want to show a date on the form and not a datetime, so we had no need for the time, but it would still toss it into the field and cause issues if the `value` attribute had quotes or not (which is just weird behavior). So this PR adds a helper function for outputting dates to form fields that makes it a lot simpler and in the correct format.

Try creating/editing a contest and messing with the sender email and sender name fields.

Screenshots of the weirdness with the inputs:

Without quotes, the time gets appended inside the input and just hangs there all weird like:
screen shot 2016-04-01 at 1 31 55 pm

With quotes, need to change the `date` type to a `datetime` for it to populate the field and show up, but we didn't want the time.
screen shot 2016-04-01 at 1 32 04 pm

Shout out to @sbsmith86 for the help in talking through the weirdness with the date input fields to help figure out a solution :dancer: 

Fixes #158

---
@DoSomething/gladiator

### DIFF
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -50,7 +50,6 @@ class ContestsController extends Controller
      */
     public function store(ContestRequest $request)
     {
-        // dd($request->all());
         $contest = Contest::create([
             'campaign_id' => $request->input('campaign_id'),
             'campaign_run_id' => $request->input('campaign_run_id'),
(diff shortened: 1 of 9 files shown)


angaither:
kewl :ship:

PR #169, opened 2016-04-01
Update to allow seeding messages and also fix issue with helper function.

weerd:
#### What's this PR do?
This PR adds a seeder for seeding Messages into the database, only for Contests that are missing them.

Try running:

```
artisan db:seed --class="MessageTableSeeder"
```
Then check the database to make sure a contest that previously was missing messages now has them.

I introduced a bug in #167 when creating a contest from scratch. While the new helper function for form dates is useful, if there is no `$contest` variable, the app would crap out. Did a temporary fix for now, but would like to figure out a better solution. Likely use the helper function in the controller before the `$contest` variable is passed and then go back to just using `{{ ... or old(...) }}` approach.

---
@DoSomething/gladiator

### DIFF
--- a/resources/views/contests/partials/_form_contest.blade.php
+++ b/resources/views/contests/partials/_form_contest.blade.php
@@ -1,11 +1,11 @@
 <div class="form-item -padded">
     <label class="field-label" for="signup_start_date">Signup Start Date:</label>
-    <input type="date" name="signup_start_date" id="signup_start_date" class="text-field" value="{{ format_date_form_field($contest->waitingRoom, 'signup_start_date') }}"></input>
+    <input type="date" name="signup_start_date" id="signup_start_date" class="text-field" value="{{ isset($contest) ? format_date_form_field($contest->waitingRoom, 'signup_start_date') : '' }}"></input>
 </div>
 
 <div class="form-item -padded">
     <label class="field-label" for="signup_end_date">Signup End Date:</label>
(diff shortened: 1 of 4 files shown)


resources/views/contests/partials/_form_contest.blade.php
```diff
@@ -1,11 +1,11 @@
 <div class="form-item -padded">
     <label class="field-label" for="signup_start_date">Signup Start Date:</label>
-    <input type="date" name="signup_start_date" id="signup_start_date" class="text-field" value="{{ format_date_form_field($contest->waitingRoom, 'signup_start_date') }}"></input>
+    <input type="date" name="signup_start_date" id="signup_start_date" class="text-field" value="{{ isset($contest) ? format_date_form_field($contest->waitingRoom, 'signup_start_date') : '' }}"></input>
 </div>
 
 <div class="form-item -padded">
     <label class="field-label" for="signup_end_date">Signup End Date:</label>
-    <input type="date" name="signup_end_date" id="signup_end_date"  class="text-field" value="{{ format_date_form_field($contest->waitingRoom, 'signup_end_date') }}"></input>
+    <input type="date" name="signup_end_date" id="signup_end_date"  class="text-field" value="{{ isset($contest) ? format_date_form_field($contest->waitingRoom, 'signup_end_date') : '' }}"></input>
```
sbsmith86: I agree with a helper in the controller, might be the best option.

sbsmith86:
:100:

PR #172, opened 2016-04-04
Cool tip

### LINKED ISSUE #146: Create a pro-tip field on leaderboard emails
As an admin, when I send out leaderboard emails, I want an optional pro-tip field

- This should appear in the email only if the field has been filled out
- There should be a different pro-tip/per leaderboard

sbsmith86:
#### What's this PR do?
Adds the pro tip field to leaderboard message creation and editing.

Pull down, edit a contest's messages and add or edit a protip. Also try creating a new contest and adding pro tips to the leaderboard messages. 

I think there could be a nicer way of making sure that messages that don't have a pro_tip field have that field set as null. Maybe a validation step that takes a message request and adds the pro_tip field as null for all the message types that don't have it. But for now, I just have the `Correspondence` class check if a `pro_tip` came through, if so, save it to the message, if not set it as null. 

Fixes #146


angaither:
will also have to make sure this is included in the body of the email if the tip is set

angaither:
:+1: for meow

PR #174, opened 2016-04-05
Start sending to users

### LINKED ISSUE #152: Actually send the right messages

sbsmith86:
#### What's this PR do?

Reworked this quite a bit...do you ever have one of those moments right before creating a PR where you are like...dafuq did I even do here :question: :exclamation: 

Here is how the logic flows from when the admin hits "send" to the user getting a message.

* Admin hits "send"
* A request is sent to the `sendMessage()` method in the `MessageController`
  * This method is responsible for grabbing all the stuff that we need to build an `Email`, instantiating the email, and firing the event that will eventually handle the message sending.
* An `Email` object gets built.
  * This `Email` object contains everything we need to send a message to multiple users.
     * It defines the tokens to look for and how to replace them
     * It creates a custom message per user. I do this here because we might need to pass in user info to the subject, and the body of the email. So this way we get a processed message for each user and we can use that to send.
     * the final object will look something like this:
```
Email {
 contest -> Contest instance (the contest the message is for)
 allMessages -> [
   0 => ['user' => Northstar User Object, 'message' => Processed Message  ]
 ]
}
```

* the Message controller fires an event and passes the `Email` to it
* The event is queued.
* When the listener runs, it pulls in the `Email`, loops through each message and queues it to be sent. All of the configuration (from, to, etc) is done here.

tell us, how can we check that it works?
Pull down branch

**VERY IMPORTANT** 
switch your `MAIL_DRIVER` to `log` so you don't really send to anyone.

run the queue listener: `php artisan queue:listen`

Kick off a send. 
check your logs, you should see an email for each person in a competition, and tokens should be replaced.

Closes #152


app/Http/Controllers/MessagesController.php
```diff
@@ -77,16 +86,28 @@ public function show($id)
      *
      * @param int $id
      */
-    public static function sendMessage(Message $message)
+    public function sendMessage(Message $message)
     {
+        // @TODO - is there a better way of bringing in this repository to this class?
+        $userRepository = app(\Gladiator\Repositories\UserRepositoryContract::class);
```
DFurnes: Can this use the `$this->userRepository` on the class now that it's not static?

app/Http/Utilities/Email.php
```diff
@@ -2,17 +2,120 @@
...
+     *
+     * @param  object $user
+     * @return array $tokens
+     */
+    protected function defineTokens($user)
+    {
+        // @TODO - set defaults?
+        $tokens = [
+            ':competition_end_date' => $this->competition->competition_end_date,
+            ':leaderboard_msg_day' => $this->competition->leaderboard_msg_day,
```
angaither: we will also want `:leaderboard_msg_day-1` in here. and other vars from phoenix as well. maybe leave #153 open?

app/Http/Controllers/MessagesController.php
sbsmith86: i don't know...honestly..im having many issues with this. lol

app/Http/Utilities/Email.php
sbsmith86: ah yea...whatever you think is best! i was thinking there would be another issue for when we need to add more tokens. I thought #153 was just for discussion since it was just labeled as a question

app/Http/Utilities/Email.php
```diff
@@ -2,17 +2,120 @@
...
+    }
+
+    /**
+     * Builds the email object.
+     */
+    protected function setupEmail()
+    {
+        // Each user gets it's own processed message
+        foreach ($this->users as $key => $user) {
+            $this->allMessages[$key]['user'] = $user;
```
angaither: i'm wondering if (since _most_ message body does not contain user names) if we could prepare the body on it's own function so we don't have to do that for each user. 
`¯\_(ツ)_/¯ `

app/Http/Utilities/Email.php
angaither: ahh, yeah i'll update it to include all

app/Http/Utilities/Email.php
sbsmith86: yea i noticed that but wanted to leave it open to the possibility...i mean if any message has user content in it, then this will need to happen (or we can like scan for the existence of a user token and only process it here if we need, but that seemed extra to me) . Or we can make the decision that user content can't be in the message body.

resources/views/messages/welcome.blade.php
```diff
@@ -1 +1 @@
-<?php echo $email->message->body ?>
+<?php echo $body; ?>
```
DFurnes: This is not a big deal, but is there any reason this isn't using Blade tags? `{!! $body !!}`

app/Listeners/QueueMessage.php
```diff
@@ -31,18 +31,22 @@ public function __construct(Mailer $mail)
...
+            $user = $message['user'];
+            $type = $content->type;
+
+            if ($user->email) {
+                $this->mail->queue('messages.' . $type, ['body' => $content->body], function ($msg) use ($email, $content, $user) {
+                    // Pulled from the contest.
+                    $msg->from($email->contest->sender_email, $email->contest->sender_name);
+
+                    // Send to user.
+                    // CAREFUL! Make sure you are using dev settings or you could actually send an email.
```
DFurnes: It might be good to short-circuit this method if `APP_DEBUG === true` just to be safe.

resources/views/messages/welcome.blade.php
sbsmith86: no i just haven't gotten around to dealing with these templates in a real way. just wanted to get the system in place...theses templates will need some of their own work cause we will probably be injecting more stuff to them and will wanna lay it out.

app/Http/Utilities/Email.php
angaither: haha yeah does seem a little extra. we can always find out if this is doing the most if it takes a year to hit and send messages to 300ish users

app/Listeners/QueueMessage.php
angaither: +1

sbsmith86: ah good idea!

sbsmith86:
@angaither @DFurnes updated:

* Pulling in UserRepository into the constructor and using it that way (`$this->userRepository`)
* Checking for `APP_DEBUG` to determine who to send to, if it is dev mode, then is sends to the user who created the contest.

app/Listeners/QueueMessage.php
```diff
@@ -31,18 +31,26 @@ public function __construct(Mailer $mail)
...
+            $type = $content->type;
+
+            if ($user->email) {
+                $this->mail->queue('messages.' . $type, ['body' => $content->body], function ($msg) use ($email, $content, $user) {
+                    // Pulled from the contest.
+                    $msg->from($email->contest->sender_email, $email->contest->sender_name);
+
+                    // Send to user.
+                    // If on testing environment, send emails to the contest creator as a test.
+                    if (! env('APP_DEBUG')) {
```
angaither: this is awesome

angaither:
looks :ok: to me! we can merge, set `APP_DEBUG` to true on QA (I think it's set to `false` currently) and start taking a look at this!!!

PR #176, opened 2016-04-06
Pull from phoenix

### LINKED ISSUE #14: Pull campaigns info from phoenix
As an admin I want these variables to be replaced in emails
- `campaign title`
- `campaign_link#prove`
- `repoortback_noun`
- `reportback_verb`

sbsmith86:
#### What's this PR do?

Pull campaign info from phoenix into gladiator.

Many TODOs here:

* store campaigns in cache and grab info from there.
* Bulk grab a set of campaigns from phoenix for pages like the contest.index view 

pull down, any where you used to see an campaign ID, you should see a real campaign title

Need to get this in so we can start pulling in campaign content into emails. There is still much to do in terms of getting this code right. 

Fixes #14

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -47,6 +47,7 @@ class CompetitionsController extends Controller
     public function show(Competition $competition)
     {
         $contest = Contest::find($competition->contest_id);
+        $campaign = $this->manager->getCampaign($contest->campaign_id);
 
         $users = [];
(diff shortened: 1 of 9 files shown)


angaither:
:+1: 
I wonder if `campaign_id` accessor should just return the title?

sbsmith86:
@angaither yea! i think something like that or... something like this https://laravel.com/docs/5.2/eloquent-serialization#appending-values-to-json...that way we can have both campaign ID (for linking and such) and title for display

angaither:
oolala :100:

PR #177, opened 2016-04-06
Updates the default messages

### LINKED ISSUE #165: Update the default messages
Since messaging isn't set yet we need to update the defaults [here](https://github.com/DoSomething/gladiator/blob/0514a20537a0d07a2edfb08706c3989ebaacbf6f/app/Http/Utilities/Correspondence.php)

deadlybutter:
#### What's this PR do?
Updates the default messages based on the [new token structure](https://github.com/DoSomething/gladiator/issues/165#issuecomment-206382375) and the [finalized message copy](https://drive.google.com/drive/u/0/folders/0B6YR3aAeso1EQ0lMR3pEcjdnblk). 

When you make new contest, do the default messages look right?
Did I make a typo in a word somewhere?

(I just read through all of them twice but could've missed something!)

🍰 

Fixes #165


app/Http/Utilities/Correspondence.php
```diff
@@ -14,72 +14,80 @@ class Correspondence
             'type' => 'welcome',
             'key' => '1',
             'label' => 'Welcome',
-            'subject' => 'Welcome, [First Name]!',
-            'body' => "Woohoo, so glad you signed up for my [Campaign Name] competition, [First Name]! Feel free to get a head start on the competition and begin today. \n\rI will be in touch with you with more info in the next few days. Excited to watch you climb the leaderboard and crush the competition! \n\r[Name of Staff]",
+            'subject' => 'Welcome, :first_name!',
+            'body' => "Woohoo, so glad you signed up for my :campaign_name competition, :first_name! Feel free to get a head start on the competition by working towards the greatest number of :reportback_noun :reportback_verb as you can. \n\rI will be in touch with you with more info in the next few days. Excited to watch you climb the leaderboard and crush the competition! \n\r:staff_name",
```
angaither: not sure if @sbsmith86 is still using the Database names but the `:staff_name` is actually stored as `:sender_name`
my mistake for not changing that in the issue

sbsmith86:
Oh i have been using whatever is in that issue. I can update that though, cause it hasn't been merged in yet.

app/Http/Utilities/Correspondence.php
```diff
@@ -14,72 +14,80 @@ class Correspondence
...
+            'subject' => '1st :campaign_name Leaderboard Update',
+            'body' => "Hello Competitors, \n\rHere is your first official :campaign_name competition update! The competition ends on :end_date. Finish in the top 3 in :reportback_noun :reportback_verb and win: \n\r- 1st place: ­ $100 amex card\r- 2nd place: ­ $50 amex card \r- 3rd place: ­ $25 amex card \n\rTwo more weeks to increase your number of :reportback_noun :reportback_verb to move up the leaderboard! \n\rPro tip ­- :pro_tip \n\rNext :reportback_noun will be due :leaderboard_msg_day-1 before 10pm est. \n\rHere is the leaderboard! Top 3 and a shoutouts below: \n\r[Leaderboard] \n\r[1 Random Image w/ comment] \n\r[Top 3 competitor IMAGES w/comment] \n\r[#] more weeks to make your mark and climb up the leaderboard!",
```
angaither: let's remove the [Leaderboard] stuff from the default message till we have a solution for how we want to build this out

deadlybutter:
I can replace all :staff_name with :sender_name if thats what we're doing

app/Http/Utilities/Correspondence.php
deadlybutter: The entire leaderboard update, or just those lines? Also I have an idea for the future for how this could work if youre interested

angaither:
@deadlybutter @sbsmith86 yeah let's do `:sender_name` since it's the same everywhere

deadlybutter:
done

app/Http/Utilities/Correspondence.php
angaither: this part, since it's likely not what we will be sending. 
`[Leaderboard] \n\r[1 Random Image w/ comment] \n\r[Top 3 competitor IMAGES w/comment] `

angaither:
oh hm. I also didn't see `\n\r[#] more weeks to make ` in the copy

deadlybutter:
@angaither in the doc there is a comment from Fantini? saying that this will just be updated manually by the editor?

app/Http/Utilities/Correspondence.php
deadlybutter: done, i think the comment is on the wrong line so it didnt fold

deadlybutter:
is this good to be merged?

angaither:
:+1:

PR #182, opened 2016-04-07
Update appending campaign data

weerd:
#### What's this PR do?
This PR updates the approach for how campaign data is appended to instances of `Contest` classes retrieved as a collection or as single model instances. It provides an easy way to send off a contest collection or model to a method on the `Manager` class called `appendCampaign()` that handles adding the full campaign data as a property on the `Contest` object. The property contains the full campaign object, so other data on the campaign is easily accessible :) If there was an issue retrieving the campaign, then the `campaign` property is still set on the contest object, but the value is set to `null`.

Load the site and see if campaign titles are being properly added on the contest's index and show pages, as well as the competition show page.

This PR provides the first step in setting up the approach for next caching the campaign data for more efficient retrieval on subsequent requests.

Refs #178

---

### DIFF
--- a/app/Services/Phoenix/Phoenix.php
+++ b/app/Services/Phoenix/Phoenix.php
@@ -24,21 +24,27 @@ class Phoenix extends RestApiClient
     }
 
     /**
-     * Send a GET request to return a campaign.
+     * Send a GET request to return all campaigns matching a given query.
      *
-     * @param  int $ids
-     * @return object
+     * @param  array  $params
+     * @return object|null
      */
-    public function getCampaign($id)
+    public function getAllCampaigns($params = [])
     {
-        $response = $this->get($this->base_uri . 'campaigns/' . $id);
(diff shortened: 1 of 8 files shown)


app/Services/Phoenix/Phoenix.php
```diff
@@ -24,21 +24,27 @@ public function __construct()
...
+    /**
+     * Send a GET request to return a campaign with the specified id.
+     *
+     * @param  string $id
+     * @return object|null
+     */
+    public function getCampaign($id)
     {
-        $response = $this->get('campaigns', $params);
+        $response = $this->get($this->base_uri . 'campaigns/' . $id);
```
weerd: GitHub is making it look like I changed a bunch of stuff here, but I didn't other than alphabetize the method name ordering.

sbsmith86:
awesome! :+1:

PR #184, opened 2016-04-08
Adds leaderboard v1

### LINKED ISSUE #97: Leaderboard
Columns for leaderboard:

Rank | Name | Number of x's y'd | Email | Flagged Status
| --- | --- | --- | --- | --- |
1 | Jerome | 50 | jeromoe@example.com | approved
2 | Hils | 35 | hils@clinton.com | approved
3 | Bernie | 20 | bernie@sanders.com | approved

Sort by number of x's y'd.

If there is a pending file highlight the row in yellow, only show approved if there are NO PENDING files

Drop phone number

deadlybutter:
#### What's this PR do?
Adds the first iteration of the leaderboard

Does it properly sort everyone based on reportback quantity?
Are all users & there details present?
Does it assign the proper ranks?

Haven't tested this with a few hundred people, not sure if it will die

Also this doesn't get signups/reportbacks as a group. Since that was proving to be difficult I'll tackle that next, but at least we are getting leaderboards live.

Fixes #97

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -46,7 +46,6 @@ class CompetitionsController extends Controller
      */
     public function show(Competition $competition)
     {
-        $this->manager->getLeaderboard($competition, $competition->users);
         $contest = Contest::find($competition->contest_id);
 
         $contest = $this->manager->appendCampaign($contest);
@@ -137,5 +136,18 @@ class CompetitionsController extends Controller
         $messages = Message::where('contest_id', '=', $contest->id)->get();
 
         return view('messages.show', compact('messages', 'competition'));
+    }
+
+    /**
+     * Display the specified leaderboard.
+     *
+     * @param  \Gladiator\Models\Competition  $competition
+     * @return \Illuminate\Http\Response
+     */
+    public function leaderboard(Competition $competition)
+    {
+        $leaderboard = $this->manager->getLeaderboard($competition, $competition->users);
+
+        return view('competitions.leaderboard', compact('competition', 'leaderboard'));
     }
 }
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -46,6 +46,7 @@ class CompetitionsController extends Controller
      */
     public function show(Competition $competition)
     {
+        $this->manager->getLeaderboard($competition);
         $contest = Contest::find($competition->contest_id);
         $campaign = $this->manager->getCampaign($contest->campaign_id);
 
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -46,7 +46,7 @@ class CompetitionsController extends Controller
      */
     public function show(Competition $competition)
     {
-        $this->manager->getLeaderboard($competition);
+        $this->manager->getLeaderboard($competition, $competition->users);
         $contest = Contest::find($competition->contest_id);
         $campaign = $this->manager->getCampaign($contest->campaign_id);
 
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -146,7 +146,7 @@ class CompetitionsController extends Controller
      */
     public function leaderboard(Competition $competition)
     {
-        $leaderboard = $this->manager->getLeaderboard($competition, $competition->users);
+        $leaderboard = $this->manager->getLeaderboard($competition);
 
         return view('competitions.leaderboard', compact('competition', 'leaderboard'));
     }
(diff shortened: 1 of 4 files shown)


deadlybutter:
@weerd ready for review!

app/Http/Controllers/CompetitionsController.php
```diff
@@ -137,4 +137,17 @@ public function message(Competition $competition, Contest $contest)
...
+
+    /**
+     * Display the specified leaderboard.
+     *
+     * @param  \Gladiator\Models\Competition  $competition
+     * @return \Illuminate\Http\Response
+     */
+    public function leaderboard(Competition $competition)
+    {
+        $leaderboard = $this->manager->getLeaderboard($competition, $competition->users);
```
weerd: I think you can call something like 
```php
$competition = $competition->users();
```
so then you can just pass `$competition` to the `getLeaderboard()` method instead of passing 2 vars, one of which is just an item of the other.

app/Services/Manager.php
```diff
@@ -92,6 +92,77 @@ public function exportCSV($model, $hasReportback = false)
...
+     * @param Competition $competition
+     * @param User $users
+     * @return array $leaderboard
+     */
+    public function getLeaderboard($competition, $users)
+    {
+        $rows = [];
+
+        // Get all users in bulk
+        $parameters['ids'] = implode(',', $users->pluck('id')->all());
```
weerd: hmmm, you can just save this as `$userIds` or something. Only reason I saved it as `$parameters['ids']` in the code you're referencing this from is cause of how it got passed in to the API as URL params.

Here you're creating an array of an array of ids, instead of just an array of ids...

app/Services/Manager.php
```diff
@@ -92,6 +92,77 @@ public function exportCSV($model, $hasReportback = false)
...
+        $users = $this->northstar->getAllUsers($parameters);
+        $users = collect($users)->keyBy('id')->all();
+
+        foreach ($users as $user) {
+            // For each user get the reportback details
+            $reportback = $this->getUserActivity($user->id, $competition);
```
angaither: isn't this going still the slow way of grabbing each user one at a time that's causing timeouts currently?

angaither: ahh I see your note now 
> Also this doesn't get signups/reportbacks as a group. Since that was proving to be difficult I'll tackle that next, but at least we are getting leaderboards live.

app/Http/Controllers/CompetitionsController.php
deadlybutter: oh sorry i thought earlier you said to pass the users from the controller

app/Services/Manager.php
weerd: We need to fix up the NS/Phoenix route for the RB retrieval to allow doing it in bulk based on user ID. Chatted about it w/ @deadlybutter for a bit. We could fix up the signups endpoint or add passing users to the `/reportbacks` endpoint since all these user's have already signed up for the campaign and all we really need are their reportbacks.

app/Services/Manager.php
deadlybutter: If I pass an array of ID's into the getAllUsers how does it know its an array of ID's? in the code we're referencing the $parameters var was just an array of ID's like it is here https://github.com/DoSomething/gladiator/blob/77ca2e1102d8af1a2878fbe79c1d56eb9f054c01/app/Services/Manager.php#L199

app/Services/Manager.php
deadlybutter: yar, doing this next, just wanted leaderboards live so ppl could see that & use it in emails or whatever.

app/Http/Controllers/CompetitionsController.php
weerd: Gotcha. Just meant to grab them in controller and pass them along as one built out larger object w/ all the info inside of it.

app/Services/Manager.php
weerd: Oh sorry. I didn't see that correctly. User the `UserRepository` with it's `getAll()` method, where you just pass an array of ids. In the code you reference, I use the Phoenix API cause for campaigns, they aren't cached currently. But user's should be via the Repository. If they're not cached, then the repository handles sending the call to Northstar. You're eschewing the use of the cache :wink:

deadlybutter: OH

deadlybutter:
@weerd updated to use the repository & also removed the users function parameter, just defined it at the top

app/Services/Manager.php
```diff
@@ -92,6 +92,77 @@ public function exportCSV($model, $hasReportback = false)
...
+    private function rankLeaderboard($leaderboard)
+    {
+        $increment = 1;
+        $rank = 1;
+
+        foreach ($leaderboard as $index => $row) {
+            // Don't perform this logic on the first element
+            if ($index > 0) {
+                // If the last row quantity equals this rows quantity, just increment.
+                if ($row['quantity'] == $leaderboard[$index - 1]['quantity']) {
```
weerd: Not a huge deal but aiming to user `===` in our PHPz!

weerd:
Quick comment, but not vital for merge. :+1:

PR #186, opened 2016-04-08
Add more tokens

sbsmith86:
#### What's this PR do?

* Added more tokens that can be used in emails:
```
            :end_date:
            :leaderboard_msg_day:
            :leaderboard_msg_day-1: 
            :first_name:           
            :sender_name:          
            :campaign_title:        
            :reportback_noun:     
            :reportback_verb:
```

**OTHER UPDATES**
* Only sends one test message to the contest sender on test environments
* Updated the leaderboard_msg_day accessor so that it just returns and `int` we can use to do that leaderboard msg logic with. 
* Moved the processing of messages into the queue handle to help with performance.

pull down branch

make sure `APP_DEBUG=true` 

Add the tokens in this PR to an email and send

You should get an email with the tokens replaced.

Addresses #153


sbsmith86:
Oh optionally, to avoid that distinct token thing I mentioned, we could have our tokens have a start and end delimiter like `:leaderboard_msg_day:`

sbsmith86:
Example email: 

image

angaither:
@sbsmith86 i'm down with having :word: ending things, we would just need to update the default messages

angaither:
woooo :+1:

PR #191, opened 2016-04-08
Proveit link token

sbsmith86:
#### What's this PR do?
Forgot the actual `:prove_it_link:` token in my last [PR]( #186 ) 

But I did add the ability to do markdown like links so now we can do things like:

`[:campaign_title:](:prove_it_link:)`

More indepth PR on the token handling is [here]( #186 )

Addresses #153


weerd:
Noice! 
:sheep: it!

PR #200, opened 2016-04-11
HOT FIX FOR EXPORTS

deadlybutter:
#### What's this PR do?
Get user data in bulk from repository
Write users in bulk to csv
Add no RB data option

does it export the things right

see slacks

Fixes # 🍰

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -9,6 +9,7 @@ use Gladiator\Repositories\UserRepositoryContract;
 use Gladiator\Services\Manager;
 use Gladiator\Models\Message;
 use Gladiator\Models\User;
+use Illuminate\Http\Request;
 
 class CompetitionsController extends Controller
 {
@@ -144,9 +145,9 @@ class CompetitionsController extends Controller
      * @param  \Gladiator\Models\Competition  $competition
      * @return \Illuminate\Http\Response
      */
-    public function leaderboard(Competition $competition)
+    public function leaderboard(Competition $competition, Request $request)
     {
-        $leaderboard = $this->manager->getLeaderboard($competition);
+        $leaderboard = $this->manager->getLeaderboard($competition, $request->input('hasReportback'));
 
         return view('competitions.leaderboard', compact('competition', 'leaderboard'));
     }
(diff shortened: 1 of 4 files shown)


weerd:
```
$csv = $this->manager->exportCSV($competition, true);
```
Isn't this the line that needs to get changed to be false?

deadlybutter:
@weerd sorry realized what you meant, updated now :D

deadlybutter:
and export without reportback data is like super fast in comparison to now

app/Http/Controllers/CompetitionsController.php
```diff
@@ -104,9 +105,9 @@ public function destroy(Competition $competition)
      * @param  \Gladiator\Models\Competition  $competition
      * @return \League\Csv\ $csv
      */
-    public function export(Competition $competition)
+    public function export(Competition $competition, Request $request)
     {
-        $csv = $this->manager->exportCSV($competition, true);
+        $csv = $this->manager->exportCSV($competition, $request->input('hasReportback'));
```
weerd: maybe use a ternary to set as `false` if nothing there instead of `null`?

deadlybutter: we can if you think it looks better, but functionally it doesnt make a difference

app/helpers.php
```diff
@@ -11,10 +11,7 @@
 function build_csv($data)
 {
     $csv = \League\Csv\Writer::createFromFileObject(new \SplTempFileObject());
-
-    foreach ($data as $row) {
-        $csv->insertOne($row);
-    }
+    $csv->insertAll($data);
```
weerd: mind adding a space above :) nit-pick :diego:

app/Http/Controllers/CompetitionsController.php
weerd: actually you're specifically passing it in in both cases. Guess it's fine for now.

deadlybutter:
added line above, i think thats what you meant

deadlybutter:
that addition caused style CI to fail?

weerd:
@deadlybutter it's not the additional line... it's the empty tab in the space that caused it to fail.

deadlybutter:
@deadlybutter install editorconfig note to self

@weerd fixed!

weerd:
👍

PR #201, opened 2016-04-11
Locking to older version of Forge until we switch to webpack.

weerd:
#### What's this PR do?
This PR locks the version of Forge we use for Gladiator to `6.6.0` and minor patches, because the new versions switch up the build system used and is more specific to the use of Webpack. So until we upgrade the build system we'll stick to a less fancy version so shit don't break!

Does shit break? No? Great. Have a 🍪 

Fixes #:cookie:

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "gulp": "^3.8.8"
   },
   "dependencies": {
-    "@dosomething/forge": "^6.6.2",
+    "@dosomething/forge": "~6.6.0",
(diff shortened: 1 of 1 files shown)


DFurnes:
:lock_with_ink_pen: :+1:

deadlybutter:
can you please add more comments?

weerd:
@deadlybutter sure....  _more comments_ :trollface:

Issue #180, opened 2016-04-07
Update RestApiClient for BOTH Northstar & Phoenix


weerd:
Right now, the `RestApiClient` class really works mostly for Northstar API requests, and it just _happens_ to also work for the most part with Phoenix API requests. But this isn't universally the case, _especially_ when trying to handle error responses... Errors from Phoenix will throw exceptions because the `RestApiClient` needs to be updated to be more generic so it applies to both APIs and the specifics are handled in the respective client service classes.

![image](https://cloud.githubusercontent.com/assets/105849/14366234/cd602d4a-fcdf-11e5-9cf9-eb2e60c2d1d9.png)

weerd:
Additional notes... this call returns something:

```
https://staging.beta.dosomething.org/api/v1/campaigns/504
```

which is a not found page on the site:

which on the staging site seems to be a "fact" and not a campaign... and the API should not be returning that as a result... so something broke in the campaign response @_@

weerd:
@sbsmith86 the above would explain some of the weird behavior w/ some campaigns working and some not. For some reason, the campaigns endpoint is now returns 11 facts, facts as appropriate responses for a queried node!

PR #202, opened 2016-04-11
Campaign data caching & general cache improvements

### LINKED ISSUE #178: Phoenix cache
cache campaign titles

weerd:
#### What's this PR do?
This PR expands on the work done in #182 to now also cache the campaign data in Redis and prefix it with `campaign:` just in case to avoid any collisions with future caching of data. The update also refactors the caching system a bit to use a `CacheStorage` trait for all the general caching methods that not _both_ the `CacheCampaignRepository` AND `CacheUserRepository` both use. DRY that shiz up! Both classes `use` the new trait to expand their functionality. 

It also breaks out a new general `RepositoryContract` for all Repository related classes, that other contracts can extend and all Repository related classed must implement.

Lastly, this PR also updates the `Contest` seeding to use real, actual campaign from DS staging. 🎉  The random numbers we were using to temporarily fill in those database fields were starting to cause too many issues with building out features, and had to constantly manually edit the ids to real campaign ids in the database. No more of that sillyness.

The usual, pul down, `dump-autoload`, do a full `migrate:refresh` and `db:seed` and then peruse the contests? Is stuff broken? Are you seeing actual campaign titles. You can also jump into Redis on Homestead and check if stuff is being cached using:

```
redis-cli
```
```
keys *
```

If you are seeing keys that look like `laravel:campaign:955` (the id number will vary) then it's working!

what else?

Fixes #178

---
@DoSomething/gladiator

### DIFF
--- a/app/Repositories/CacheCampaignRepository.php
+++ b/app/Repositories/CacheCampaignRepository.php
@@ -2,8 +2,6 @@
 
 namespace Gladiator\Repositories;
 
-use Gladiator\Repositories\CacheStorage;
-
 class CacheCampaignRepository implements RepositoryContract
 {
     use CacheStorage;
--- a/app/Repositories/CacheCampaignRepository.php
+++ b/app/Repositories/CacheCampaignRepository.php
@@ -26,7 +26,7 @@ class CacheCampaignRepository implements RepositoryContract
      */
     public function find($id)
     {
-        $key = $this->setPrefix($id, $this->prefix);
+        $key = $this->setPrefix($id);
 
         $campaign = $this->retrieve($key);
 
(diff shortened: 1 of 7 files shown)


app/Repositories/CacheCampaignRepository.php
```diff
@@ -0,0 +1,51 @@
...
+     * @return object|null
+     */
+    public function find($id)
+    {
+        $key = $this->setPrefix($id, $this->prefix);
+
+        $campaign = $this->retrieve($key);
+
+        if (! $campaign) {
+            // @see https://github.com/DoSomething/gladiator/issues/180
```
weerd: _Reminder:_ review this section when addressing #180

app/Repositories/CacheStorage.php
```diff
@@ -0,0 +1,145 @@
+<?php
+
+namespace Gladiator\Repositories;
+
+use Illuminate\Support\Facades\Cache;
+
+trait CacheStorage
```
sbsmith86: Sorry, can you briefly explain why we need our own implementations of these methods? are there business logic reasons to define say a custom implementation of `Cache::flush()` or `Cache::get()`

database/factories/ModelFactory.php
```diff
@@ -25,9 +26,16 @@
 
 // Contest Factory.
 $factory->define(Contest::class, function (Generator $faker) {
+
+    $phoenix = new Phoenix;
+
+    $campaignIds = ['1247', '1663', '1581', '1593', '1467', '1334', '362', '46', '1650', '1560', '1222', '1667', '74', '1198', '886', '1492', '850', '1503', '1376', '1665', '955', '31'];
```
sbsmith86: YASSS :clap:

app/Services/Manager.php
```diff
@@ -303,7 +305,9 @@ protected function appendCampaignToCollection($collection)
      */
     protected function appendCampaignToModel($model)
     {
-        $campaign = $this->phoenix->getCampaign((string) $model->campaign_id);
+        $campaignId = (string) $model->campaign_id;
```
sbsmith86: this doesn't already comeback from the model as a string?

app/Repositories/CacheStorage.php
weerd: @DFurnes brought up a similar point when I did the user cache stuff. While not necessarily needed at the moment, since we're doing our own implementations of some of the others that `Cache` methods that involve a bit of custom business logic for our app, I wanted to add these in here as well so that they're all in one place. Then we only really need to consider importing the `Cache` facade in one location and always know where to go to troubleshoot stuff if something's not working, instead of find their use scattered around in the separate `Cache*Repository` classes. More of a convenience thing for now, and if it turns out we do need to expand them w/ extra business logic, they're at the ready :)

sbsmith86: ok i see. cool! thanks!

weerd: Sorry, that probs wasn't brief 😬

app/Services/Manager.php
weerd: Not that I was noticing no. Wanted to cast it just in case.

app/Repositories/CacheStorage.php
sbsmith86: @weerd yet, was a good explanation! no regrets!

sbsmith86:
:100: :+1:

weerd:
👯

PR #204, opened 2016-04-12
Multiuser

### LINKED ISSUE #194: Make leaderboard faster
Solve for the fact that #184 "doesn't get signups/reportbacks as a group. Since that was proving to be difficult I'll tackle that next, but at least we are getting leaderboards live."

deadlybutter:
#### What's this PR do?
- Adds ability to fetch many user signups at once
- Updates leaderboard to use that logic 

Load a leaderboard, does shit work?

this has been a real pain in the arse

Fixes #194

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -260,19 +260,19 @@ class Manager
         return null;
     }
 
-    public function getActivityForAllUsers($ids, $model)
+    public function getActivityForAllUsers($ids, $model, $batchSize = 50)
     {
         $campaign = $model->contest->campaign_id;
         $campaign_run = $model->contest->campaign_run_id;
 
         $signups = [];
-        $count = intval(ceil(count($ids) / 50));
+        $count = intval(ceil(count($ids) / $batchSize));
         $index = 0;
 
         for ($i = 0; $i < $count; $i++) {
-            $batch = array_slice($ids, $index, 50);
+            $batch = array_slice($ids, $index, $batchSize);
             $signups = array_merge($signups, $this->getUserSignup(implode(',', $batch), $campaign, $campaign_run, true));
-            $index += 50;
+            $index += $batchSize;
         }
 
         return collect($signups)->keyBy(function ($item) {
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -280,7 +280,7 @@ class Manager
             $signups = array_merge($signups, $this->northstar->getUserSignups(implode(',', $batch), $campaign, $campaign_run));
             $index += $batchSize;
         }
-        
+
         return collect($signups)->keyBy(function ($item) {
             return $item->user->id;
         });
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -204,11 +204,11 @@ class Manager
      *
      * @return object $signups
      */
-    public function getUserSignup($id, $campaign = null, $campaignRun = null, $getMultiple = false)
+    public function getAllUserSignups($ids, $campaign = null, $campaignRun = null)
     {
-        $signups = $this->northstar->getUserSignups($id, $campaign);
+        $signups = $this->northstar->getAllUserSignups($ids, $campaign);
 
-        if ($getMultiple) {
+        if (is_array($ids)) {
             $multipleSignups = [];
         }
 
@@ -216,7 +216,7 @@ class Manager
         if ($campaignRun) {
             foreach ($signups as $key => $signup) {
                 if ($signup->campaign_run->id === (string) $campaignRun) {
-                    if ($getMultiple) {
+                    if (is_array($ids)) {
                         array_push($multipleSignups, $signups[$key]);
                     } else {
                         $signups = $signup[$key];
@@ -225,7 +225,7 @@ class Manager
             }
         }
 
-        if ($getMultiple) {
+        if (is_array($ids)) {
             return $multipleSignups;
         }
 
@@ -246,7 +246,7 @@ class Manager
         $campaign = $model->contest->campaign_id;
         $campaign_run = $model->contest->campaign_run_id;
 
-        $signup = $this->getUserSignup($id, $campaign, $campaign_run);
+        $signup = $this->getAllUserSignups($id, $campaign, $campaign_run);
 
         if (is_array($signup)) {
             $signup = reset($signup);
@@ -260,6 +260,9 @@ class Manager
         return null;
     }
 
+    /**
+     *
+     */
     public function getActivityForAllUsers($ids, $model, $batchSize = 50)
     {
         $campaign = $model->contest->campaign_id;
@@ -271,7 +274,7 @@ class Manager
 
         for ($i = 0; $i < $count; $i++) {
             $batch = array_slice($ids, $index, $batchSize);
-            $signups = array_merge($signups, $this->getUserSignup(implode(',', $batch), $campaign, $campaign_run, true));
+            $signups = array_merge($signups, $this->getAllUserSignups(implode(',', $batch), $campaign, $campaign_run));
             $index += $batchSize;
         }
 
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -98,7 +98,7 @@ class Manager
      * Build leaderboard data for a given Competition
      *
      * @param Competition $competition
-     * @param User $users
+     * @param bool $ignoreFlagged Should we include flagged reportbacks?
      * @return array $leaderboard
      */
     public function getLeaderboard($competition, $ignoreFlagged = false)
@@ -230,7 +230,7 @@ class Manager
      *
      * @param array $ids Array of ID's
      * @param \Gladiator\Models\Competition $competition
-     *
+     * @param int $batchSize How many reportbacks to fetch at once
      * @return collection $signups
      */
     public function getActivityForAllUsers($ids, $competition, $batchSize = 50)
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -207,33 +207,28 @@ class Manager
      *
      * @return object $signups
      */
-    public function getAllUserSignups($ids, $campaign = null, $campaignRun = null)
-    {
-        $signups = $this->northstar->getAllUserSignups($ids, $campaign);
+     public function getAllUserSignups($ids, $campaign = null, $campaignRun = null)
+     {
+         $signups = $this->northstar->getUserSignups($ids, $campaign);
 
-        if (is_array($ids)) {
-            $multipleSignups = [];
-        }
+         $data = [];
 
-        // Only return the sign up record for the run that was specified.
-        if ($campaignRun) {
-            foreach ($signups as $key => $signup) {
-                if ($signup->campaign_run->id === (string) $campaignRun) {
-                    if (is_array($ids)) {
-                        array_push($multipleSignups, $signups[$key]);
-                    } else {
-                        $signups = $signup[$key];
-                    }
-                }
-            }
-        }
+         if ($campaignRun) {
+             foreach($signups as $key => $signup) {
+                 if ($signup->campaign_run->id === (string) $campaignRun) {
+                     array_push($data, $signups[$key]);
+                 }
+             }
 
-        if (is_array($ids)) {
-            return $multipleSignups;
-        }
+             if (sizeof(explode(",", $ids)) <= 1) {
+                 return array_shift($data);
+             }
 
-        return $signups;
-    }
+             return $data;
+         }
+
+         return $signups;
+     }
 
     /**
      * Get a user's signup/reportback activity for a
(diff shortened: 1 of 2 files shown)


app/Services/Manager.php
```diff
@@ -98,21 +98,36 @@ public function exportCSV($model, $hasReportback = false)
...
+    public function getLeaderboard($competition, $ignoreFlagged = false)
     {
         $rows = [];
         $users = $competition->users;
 
         // Get all users in bulk
-        $users = $this->repository->getAll($users->pluck('id')->all());
-        $users = $users->keyBy('id')->all();
+        $ids = $users->pluck('id')->all();
+        $users = $this->repository->getAll($ids)->keyBy('id')->all();
```
weerd: Will need to update this to `userRepository->` after I merge #202

deadlybutter: _make me_ 👊

deadlybutter: okie dokes, you should merge your stuff first then and ill resolve the conflicts

app/Services/Manager.php
```diff
@@ -98,21 +98,36 @@ public function exportCSV($model, $hasReportback = false)
...
+        $signups = $this->getUsersActivity($ids, $competition);
```
weerd: can we rename this function to `getActivityForAllUsers()`... it's deceiving in that it sounds like it's an apostrophe "s" as in "get this User's activity"

deadlybutter: yar that makes sense

app/Services/Manager.php
```diff
@@ -226,30 +253,55 @@ public function getUserActivity($id, $model)
...
-            // Return the reportback.
-            return $signup->reportback;
+            return $this->formatReportback($signup);
         }
 
         // If the user has no activity for this competition or waiting room.
         return null;
     }
 
+    public function getUsersActivity($ids, $model)
```
weerd: Might be useful to add the `$size` argument in this function like it is in https://github.com/DoSomething/gladiator/blob/00b1b5a34af89a78ff3d3bc9c6f05b31bfc7e0f8/app/Repositories/DatabaseUserRepository.php#L143

deadlybutter: i didnt entirely understand the point of it. It all gets merged into the same array at the end?

weerd: If it turns out you could pass more then you could up it based on custom logic and could mean less calls to external API.

deadlybutter: hmmm ok

app/Services/Manager.php
```diff
@@ -226,30 +253,55 @@ public function getUserActivity($id, $model)
...
+        $signup->reportback->updated_at = $signup->reportback->updated_at->format('Y-m-d');
+
+        // Set flagged status to 'pending' if it is NULL, otherwise use bool value
+        if (! isset($signup->reportback->flagged)) {
+            $signup->reportback->flagged = 'pending';
+        } elseif ($signup->reportback->flagged) {
+            $signup->reportback->flagged = 'flagged';
+        } else {
+            $signup->reportback->flagged = 'approved';
+        }
```
weerd: Just to be sure... here we only want to look at the Reportback and _not_ the Reportback Items, right?

app/Services/Manager.php
```diff
@@ -226,30 +253,55 @@ public function getUserActivity($id, $model)
...
+
+        // Return the reportback.
+        return $signup->reportback;
```
weerd: do you want to return just the reportback or the full signup? Since you pass in the full signup, to format stuff, I would assume you'd be sending the full signup back...?

weerd: If not the case, maybe _just_ pass in the reportback to begin with:

```php
private function formatReportback($reportback) { ... }
``

Would go more w/ the name of the function too ;)

File: app/Services/Manager.php (hunk 4 above)
Comment (deadlybutter) [inline app/Services/Manager.php]: I just moved the existing code into a new function so it can be called by other stuff, I assume this is how it was intended

File: app/Services/Manager.php (hunk 5 above)
Comment (deadlybutter) [inline app/Services/Manager.php]: the function is called formatReportback, I dont think it would make sense to return the signup

Comment (deadlybutter) [inline app/Services/Manager.php]: i guess it could just get the reportback passed to it and we remove the need for signup->

File: app/Services/Northstar/Northstar.php [hunk 6]
```diff
@@ -71,7 +71,7 @@ public function verifyUser($credentials)
     public function getUserSignups($id, $campaigns)
     {
         $response = $this->get('signups', [
-            'user' => $id,
+            'users' => $id,
```
weerd: plural but only 1 `$id`?

app/Services/Manager.php
weerd: sounds good.

app/Services/Manager.php
```diff
@@ -98,21 +98,36 @@ public function exportCSV($model, $hasReportback = false)
...
+        $ids = $users->pluck('id')->all();
+        $users = $this->repository->getAll($ids)->keyBy('id')->all();
+        $signups = $this->getActivityForAllUsers($ids, $competition);
 
         foreach ($users as $user) {
+            $id = $user->id;
             // For each user get the reportback details
-            $reportback = $this->getUserActivity($user->id, $competition);
+            if (! $signups->has($id)) {
+                continue;
```
weerd: This jumps us out of the `foreach` loop, right?

app/Services/Manager.php
```diff
@@ -98,21 +98,36 @@ public function exportCSV($model, $hasReportback = false)
...
+            }
+
+            $signup = $signups->get($id);
+            if (! isset($signup->reportback)) {
+                continue;
+            }
+
+            $reportback = $this->formatReportback($signup);
```
weerd: yeah, jut pass the reportback here `formatReportback ($singup->reportback)`

app/Services/Northstar/Northstar.php
deadlybutter: well users can work for both 1 id, or both. remember you had me change this in drupals?

app/Services/Manager.php
deadlybutter: no continues just skip to the next thingy

app/Services/Manager.php
deadlybutter: yar refresh the page

app/Services/Northstar/Northstar.php
weerd: k. change `$id` to `$ids`... when in doubt pluralize

app/Services/Manager.php
weerd: oh, sorry I meant jumps out and moves to the next item in loop, right?

deadlybutter: yeah

app/Services/Manager.php
```diff
@@ -226,30 +253,55 @@ public function getUserActivity($id, $model)
...
-            // Return the reportback.
-            return $signup->reportback;
+            return $this->formatReportback($signup);
         }
 
         // If the user has no activity for this competition or waiting room.
         return null;
     }
 
+    public function getActivityForAllUsers($ids, $model, $batchSize = 50)
```
weerd: missing docblocks

app/Services/Manager.php
```diff
@@ -226,30 +253,55 @@ public function getUserActivity($id, $model)
...
+            $signups = array_merge($signups, $this->getUserSignup(implode(',', $batch), $campaign, $campaign_run, true));
+            $index += $batchSize;
+        }
+
+        return collect($signups)->keyBy(function ($item) {
+            return $item->user->id;
+        });
+    }
+
+    private function formatReportback($reportback)
```
weerd: missing docblocks

app/Services/Manager.php
```diff
@@ -189,19 +204,31 @@ public function collectContestInfo($data)
...
      * @return object $signups
      */
-    public function getUserSignup($id, $campaign = null, $campaignRun = null)
+    public function getUserSignup($id, $campaign = null, $campaignRun = null, $getMultiple = false)
     {
         $signups = $this->northstar->getUserSignups($id, $campaign);
 
+        if ($getMultiple) {
+            $multipleSignups = [];
+        }
```
weerd: Instead of adding the additional `$getMultiple` param, couldn't you just do a check if the `$id` (which could be renamed to `$ids`) is an array?

deadlybutter: $signups is always an array, even if it is just 1 thing (otherwise the for each loop wouldnt work under it)

weerd: what's the multiple signups here? So this function gets a signup for a specific campaign and run for a specific user id? So is this multiple signups for different campaigns for a single user?

deadlybutter: no, the problem is originally this function would loop through all of the results, but it would only record them into one variable. thus, if i wanted return multiple signups, i cant.

$multipleSignups indicates we should return an array of results, instead of just one

sbsmith86: I think either we should rename this to `getUsersSignups()` (plural) or maybe separate out into two different functions where one returns multiple signups and one returns a single signup. From a pure readability POV it seems weird to call `getUserSignup` but have to designate that you want multiple signups. But if we just pluralize it from the jump it makes sense that if you only provide one `id` you will only get one signup.

weerd: @sbsmith86 yeah, good call. I think part of my hangup is in the naming and the numerous things it's trying to accomplish in one go.

app/Services/Manager.php
sbsmith86: Why do we need to pass the model in here? based on how you are pulling in campaign_id below, the function sorta assumes you are starting from a competition, right?

deadlybutter:
@sbsmith86 @weerd hows `getAllUserSignups` now?

app/Services/Manager.php
deadlybutter: @sbsmith86 you could pass both a waiting room or a competition here if you wanted

deadlybutter: not really sure why youd want that for a waiting room tho... i guess we can call it competition

sbsmith86: true!...i guess it doesn't hurt to allow us to call this for either waiting rooms or competitions..you never know.

deadlybutter:
@weerd ready for review!

deadlybutter:
actually @weerd -- @chloealee finished implementing the run stuff yesterday, so I ditched the signups function altogether and this is just using northstar now, much simpler

app/Services/Manager.php
```diff
@@ -183,32 +198,6 @@ public function collectContestInfo($data)
...
-     * Get user signup for a specific campaign and run. If only user id is
-     * provided, the send back all user signups.
-     *
-     * @param  string $id
-     * @param  string $campaign    Campaign ID
-     * @param  int    $campaignRun Campaign Run ID
-     *
-     * @return object $signups
-     */
-    public function getUserSignup($id, $campaign = null, $campaignRun = null)
```
deadlybutter: 💥

deadlybutter:
The werker build failed?

`SyntaxError: Unexpected token u`

👓  lookin around but not sure where this troubling U is

deadlybutter:
Yeah the Werker derp was a Werker derp. Same code built fine in a diff PR https://app.wercker.com/#build/570e64dbdf1b13444300b120

app/Services/Manager.php
```diff
@@ -101,21 +101,36 @@ public function exportCSV($model, $hasReportback = false)
      * @param User $users
      * @return array $leaderboard
```
weerd: Update the docblocks... the params changed 😉

app/Services/Manager.php
```diff
@@ -222,38 +211,77 @@ public function getUserActivity($id, $model)
...
         // If the user has no activity for this competition or waiting room.
         return null;
     }
 
     /**
+     * Get reportbacks for many users & a given competition.
+     *
+     * @param array $ids Array of ID's
+     * @param \Gladiator\Models\Competition $competition
+     *
```
weerd: no space between @param & @return

app/Services/Manager.php
```diff
@@ -222,38 +211,77 @@ public function getUserActivity($id, $model)
...
 
         // If the user has no activity for this competition or waiting room.
         return null;
     }
 
     /**
+     * Get reportbacks for many users & a given competition.
+     *
+     * @param array $ids Array of ID's
+     * @param \Gladiator\Models\Competition $competition
```
weerd: Add missing `$batchSize` param 😉

app/Services/Northstar/Northstar.php
```diff
@@ -66,13 +66,15 @@ public function verifyUser($credentials)
      *
      * @param  string $id
```
weerd: param is plural now 😉  comment 🐦

weerd:
👍

PR #209, opened 2016-04-13
Upgrades leaderboard

### LINKED ISSUE #179: Update CSV export
this should try to call users with the mutli-user endpoint instead of grabbing each user one by one

exports are timing out

deadlybutter:
#### What's this PR do?
Updates leaderboard to use batching for reportback activity 

Does the export even work?
Is it faster?

Fixes #179

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -57,13 +57,15 @@ class Manager
         $data = [];
         $users = $model->users;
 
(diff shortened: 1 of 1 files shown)


weerd:
@deadlybutter looks good 👍 

Gonna merge so I can have this code for the issue I'm about to work on 😉

PR #210, opened 2016-04-13
Send test email

### LINKED ISSUE #198: Test email functionality
We need to build out a better way for admins to test emails before sending.

Current behavior: On QA, admin clicks send and one email will be sent to the person who is designated the sender email on the contest. In prod, if and admin clicks send, it will really send this email to every person in the competition

Desired behavior: We provide a test email button that just fires off and email to the admin this way emails can be tested in QA and prod.

sbsmith86:
#### What's this PR do?

Added a "test" button to the message sending page that will send a message to the email and name set as the contest `sender_email` and `sender_name`

image

Pull down branch. Set your email as the `sender_email` in a contest. Go to send an email. Click the "test" button and you should receive an email with the tokens replaced with your user info and the correct contest/competition/campaign information.

This removes the previous need for `APP_DEBUG=true` on dev and QA environments because now we can just manually fire off a test instead of deciding who to send to based on which environment we are in.

Fixes #198


weerd:
Just from visual of the screenshot... do we need the subject & body in that listing? Would the label be enough to know what email is getting tested/sent. Removing those 2 columns could help cleanup the table a bit and make it much easier to scan. Plus the label link would show you that particular message right?

app/Http/Controllers/CompetitionsController.php
```diff
@@ -51,12 +51,7 @@ public function show(Competition $competition)
 
         $contest = $this->manager->appendCampaign($contest);
 
-        $users = [];
-        $ids = $competition->users->pluck('id')->toArray();
-
-        if ($ids) {
-            $users = $this->repository->getAll($ids);
-        }
+        $users = $this->manager->getModelUsers($competition);
```
weerd: Nice! Good call on this 💃

app/Services/Manager.php
```diff
@@ -198,6 +198,25 @@ public function collectContestInfo($data)
...
+    public function getModelUsers($model)
+    {
+        $users = [];
+        $ids = $model->users->pluck('id')->toArray();
+
+        if ($ids) {
+            $users = $this->userRepository->getAll($ids);
+        }
+
+        return $users;
```
weerd: to maybe clean this up in case `$ids` is null or a model is passed that doesn't have users...

```php
if ($ids) {
  return $this->userRepository->getAll($ids);
}

return null;
```

app/Http/Controllers/MessagesController.php
```diff
@@ -96,12 +96,15 @@ public function sendMessage(Message $message)
...
-            $users = $this->userRepository->getAll($ids);
+        // If sending test email, create a fake user object with the contest's
+        // sender_name and sender_email so we prep and send a message only for the admin.
+        if (request('test')) {
+            $users[] = (object) [
+                'first_name' => $contest->sender_name,
+                'email' => $contest->sender_email,
+            ];
+        } else {
+            $users = $this->manager->getCompetitionUsers($competition);
```
weerd: am I blind or does the `getCompetitionUsers()` method not exist on the `Manager` class?

sbsmith86: YES omg. my bad. good catch.

app/Services/Manager.php
sbsmith86: ah yea that is nicer

sbsmith86:
@weerd as far as the table on the send page goes, I'm a little indifferent, we certainly remove the subject and body columns if it isn't useful...I found it a little useful when testing to just have a quick scan of what the content for that email I should be expecting. Or like when I forgot which message I had just sent off for a test, having the subject right there helped me reference the email I just got. Lol but really up in the air on that.

weerd:
Cool. Maybe leave the subject then if it's helpful, but the body feels extraneous and gets clipped anyhoo :)

weerd:
Looks good 👍 
:sheep: it!

PR #218, opened 2016-04-14
Export improvements

deadlybutter:
#### What's this PR do?
- Removes unnecessary export option & simplifies (_removes_) conditional logic
- (Hopefully) fixes export on prod by specifying the count based on batch size.

NOTE: The second thing is a temp hack, we should integrate the northstar library which handles pagination.

Do exports still work? 
Does reportback data show on prod?

It's very hard to reproduce this problem failing, and given we've eliminated all of the other problems me, dave diego and fantini can think of, me and dave came to the conclusion it might be pagination.

Fixes #:cake:


weerd:
@deadlybutter  this is going to likely conflict w/ a bunch of stuff I'm already working on 😱

DFurnes:
Code seems fine to me. :+1:

PR #224, opened 2016-04-14
Refactoring exports and leaderboard processes to simplify and DRY code up.

### LINKED ISSUE #205: User Reportback Item status for leaderboard information
We are currently using the `/signups` endpoint to retrieve Reportbacks and Reportback items for user contestants.

The status in the leaderboard of a submission is currently based on the `flagged` status of a Reportback. However, if we want to display status of a submission, such as which ones are `promoted`, `approved`, `flagged` (different from Reportback flagging), etc, we need to be referencing the `status` property on the individual Reportback Items for a particular "parent" Reportback.

_*Side Note:_ Eventually, we may want to switch to using the `/reportbacks` endpoint and allow passing a `users` parameter to the API (update to Phoenix API needed), instead of going through the `/signups` endpoint, since we _already_ know that all these users are signed up because they are specifically _in_ a competition. So we're requesting a bunch of extraneous data and complicating the logic by having to go through the `/signups` endpoint.

weerd:
#### What's this PR do?
This PR updates the CSV exporting functionality and the leaderboard generation. It DRY's up the code and cleans up the exporting process to make it easier to troubleshoot issues. For the leaderboard it now appends retrieved user reportback activity to the user object for easier retrieval. Also, the leaderboard export functionality was moved to the leaderboard page to help avoid having to make the same series of costly API and data requests when viewing the leaderboard page, and then again when exporting the leaderboard. Instead, now when the leaderboard page is loaded, the data is flashed in cache (1 request cycle) so when exporting the data, it will grab it from that flashed cache session. This also ensures that the leaderboard you see is the one that gets exported.

The leaderboard view also now includes an additional table with the list of contestants that have yet to reportback.

#### Screenshots
pasted image at 2016_04_18 04_02 pm

Pull down, do the usual composer-y thngs. Load a competition and view the leaderboard. Then try exporting the data. Also try exporting the list of competition users on the competitions page and waiting room pages to make sure you get an accurate CSV exported.

It was a bit difficult to test the retrieval of signup activity data from the set of test users I was using, since there wasn't necessarily data associated with that user and the randomly picked campaign they were assigned to in the seeding process. A lot of manual effort needed to be made to ensure obtaining users that could be tested. There is some concern how the data will be when testing this PR on QA environments.

Fixes #205

---
@DoSomething/gladiator


app/Http/Controllers/CompetitionsController.php
```diff
@@ -94,15 +95,38 @@ public function destroy(Competition $competition)
...
-     * @return \League\Csv\ $csv
+     * @param  \Illuminate\Http\Request  $request
+     * @return void
      */
-    public function export(Competition $competition)
+    public function export(Competition $competition, Request $request)
     {
-        $csv = $this->manager->exportCSV($competition);
-        $csv->output('competition' . $competition->id . '.csv');
+        $fileName = 'contest_' . $competition->contest_id . '-' . 'competition_' . $competition->id;
```
deadlybutter: would it make more sense to replace contest id with campaign name? might be easier for chris/hillary to recognize

weerd: Potentially, might just be a really long file name though. Also using this for flash caching so it helps to have a shorter, predictable name I can check the cache for. Maybe a later upgrade though, w/ a function to switch spaces to hyphens, etc?

weerd: Might also be useful to add the campaign name to the leaderboard view? Instead of saying "Viewing leaderboard for competition 4" for example?

app/Http/Controllers/CompetitionsController.php
```diff
@@ -94,15 +95,38 @@ public function destroy(Competition $competition)
...
+
+        $withReportback = convert_string_to_boolean($request->input('withReportback'));
```
deadlybutter: and it returns, again!

app/Http/Controllers/CompetitionsController.php
```diff
@@ -94,15 +95,38 @@ public function destroy(Competition $competition)
...
+
+        $users = $this->manager->getModelUsers($competition, $withReportback);
+
+        if (! $withReportback) {
+            // @TODO: can probably flash this to the session too!
+            $fileName = $fileName . '-users';
+        } else {
+            $fileName = $fileName . '-leaderboard';
+
+            if (session()->has($fileName)) {
```
deadlybutter: just curious, why are you using session over passing parameters?

app/Http/Controllers/CompetitionsController.php
weerd: Yeah. Made sense to keep it otherwise we end up checking signups always if not.

app/Http/Controllers/CompetitionsController.php
weerd: When the leaderboard page loads that is request #1, the leaderboard view shows and we see the leaderboard, end of execution. If we then export the leaderboard as a CSV, that is request #2, it would need to go and do all the DB & API queries all over again to grab that info and then build the CSV to download (what we were doing before). Instead, I flash all that data to the session and it persists in a very very temporary cache and is available just for request #2. Super fast export of leaderboard!!

weerd: @deadlybutter I store the session here: https://github.com/DoSomething/gladiator/pull/224/files#diff-626b13c4d9542a855923a32d7ed83d74R177

app/Http/Controllers/CompetitionsController.php
deadlybutter: Mmm yeaha that makes sense

app/Http/Controllers/CompetitionsController.php
deadlybutter: that makes lots of sense. awesome 👍

app/Http/Controllers/CompetitionsController.php
angaither: @weerd i like the campaign name + comp number in the view

weerd: cool. I can add that :)

deadlybutter:
phew that was a hefty PR 👍 

giphy

weerd:
🍾 🚢

PR #230, opened 2016-04-19
Adds a way to specify the max # users per competition

### LINKED ISSUE #193: Allow admins to choose # of competitions to split a waiting room into
Right now splitting a waiting room has the following rules:

- <300 people, split into 1 competition
- >300 people, split evenly into x competitions all under 300 people

Ideally when an admin splits a waiting they could make an explicit choice of how many competitions they want to split it into.

This way if say a waiting room has 301 people, an admin can run one competition of 301 instead of two competitions of 150 and 151 respectively.

deadlybutter:
#### What's this PR do?
Adds option on competition form to specify the max users per competition.

Does the live updating work?
Does the split functionality accurately reflect the preview?

Fixes #193


package.json
```diff
@@ -5,6 +5,8 @@
   },
   "dependencies": {
     "@dosomething/forge": "~6.6.0",
+    "babel-preset-es2015": "^6.6.0",
```
deadlybutter: this had to be added for elixr to work properly. see this issue: https://github.com/laravel/elixir/issues/354

app/Http/Controllers/Api/WaitingRoomsController.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Gladiator\Http\Controllers\Api;
+
+use Gladiator\Models\WaitingRoom;
+use Illuminate\Http\Request;
+
+class WaitingRoomsController extends ApiController
+{
+    public function calculateSplit(Request $request)
```
weerd: Checkout the Controllers/Api/UsersController.php You need to use the ApiAuth middleware stuff or else anyone could do the calculate.

package.json
```diff
@@ -5,6 +5,8 @@
...
+    "babel-preset-react": "^6.5.0",
```
weerd: Are you using React?

DFurnes: From the issue @deadlybutter linked, it looks like [Elixir hardcodes them](https://github.com/laravel/elixir/blob/aac2fd11fe5b7868dfe23a30d54474bfa5c2f33f/src/Config.js#L332) in the Gulp configuration rather than letting you set them in your `.babelrc`. The transforms should be no-ops if you don't have JSX in your scripts, so no real harm in including.

app/Http/Controllers/Api/WaitingRoomsController.php
DFurnes: I think we'd actually want to do a CSRF-style middleware (if anything) here, since this API endpoint is being called from the client-side (and we defs don't want to expose our API key in our HTML templates).

package.json
weerd: Ah ok. Sorry, missed that part.

gulpfile.js
```diff
@@ -14,5 +14,6 @@ var elixir = require('laravel-elixir');
 elixir(function(mix) {
     mix.copy('node_modules/@dosomething/forge/dist', 'public/assets/vendor/forge');
     mix.copy('node_modules/dosomething-modal/dist', 'public/assets/vendor/modal');
+    mix.browserify('split.js', 'public/assets/js/split.js');
```
DFurnes: This should be a "main" entry point, like a `app.js` or `main.js`, that then imports your `split.js` module (so if we add any other scripts in the future, they can all easily be included in the same bundle).

app/Http/Controllers/Api/WaitingRoomsController.php
deadlybutter: Didn't realize this was a big deal as it only returns user ID's. I can look into adding the CSRF middleware

resources/assets/js/split.js
```diff
@@ -0,0 +1,14 @@
+$('#competition_max').change(function() {
+  var waitingRoom = window.location.pathname.split('/')[2];
```
DFurnes: Hmmmm, maybe this  is something we should be including in the page? This could break unexpectedly if we change the URL structure (like adding a prefix or something).

deadlybutter: how would we do that, embed a javascript object in the page?

app/Http/Controllers/Api/WaitingRoomsController.php
weerd: So if not really used for the Api, does it need to be a separate controller? Why not just a method in the normal controller?

weerd: I may very well just be missing a key thing? :shrug:

deadlybutter: it is used by the API, its a json call from the JS

weerd: But it's for internal stuff. Phoenix would never have to hit this endpoint. Nothing we'd add to the API documentation.

I'd rather keep stuff with controllers related to the actual API w/in the Controllers/Api...

The JS script could very well just hit a dedicated route that's part of the waitingroom and lives as a method on the regular WaitingRoom controller like:
```php
Route::post('waitingrooms/{waitingrooms}/split/calculate', 'WaitingRoomsController@splitCalculation')->name('split');
```

Or something like the above.

deadlybutter: sure I can do that!

weerd: Cool :) I think unless it's something that we'd document in our API docs etc and potentially used by external API or whatevs, let's just keep those methods to internal controllers.

resources/assets/js/split.js
weerd: maybe harness some `data-*` attributes we can pull info from?

gulpfile.js
weerd: @deadlybutter I can show you more on this when I'm in tomorrow. We'll defs be adding more scripts in the near future, so having them bundled all together as a `main.js` like @DFurnes mentions is ideal!  💃

app/Http/Controllers/Api/WaitingRoomsController.php
deadlybutter: So if we're using the internal controller, do we still care at all about API keys?

If we do care, one option to is embed both the API URL in a data attribute as @weerd suggested but ALSO embed an api key. You'd only be able to read that if you were able to access the page.

But again not really sure what the risk is here. Northstar user id's are exposed everywhere publicly, and you can't actually split this way, it just returns hypothetical match ups.

resources/assets/js/split.js
DFurnes: > maybe harness some data-* attributes we can pull info from?

👍

app/Http/Controllers/Api/WaitingRoomsController.php
DFurnes: I think for something like this [a CSRF token header](https://laravel.com/docs/5.2/routing#csrf-x-csrf-token) is the right (and probably easiest!) choice.

DFurnes: I think @weerd makes good points on the "split" between API vs "web" controllers.

weerd:
@DFurnes, @deadlybutter and I chatted IRL and decided, since it's just doing some simple math, there's no need to be querying back and forth with the DB to do the calculation. We're adding a `data-total` attribute in the form and then grabbing the total from there to do the simple splitting math in JS. Then on form submit, we submit the value the admin deems usable to the POST request to do the actual splitting and DB related actions. Clean and simple!

deadlybutter:
@weerd should be good to go now

gulpfile.js
```diff
@@ -14,5 +14,9 @@ var elixir = require('laravel-elixir');
 elixir(function(mix) {
     mix.copy('node_modules/@dosomething/forge/dist', 'public/assets/vendor/forge');
     mix.copy('node_modules/dosomething-modal/dist', 'public/assets/vendor/modal');
+    // mix.browserify('split.js', 'public/assets/js/split.js');
```
weerd: Probs can remove this right?

deadlybutter: whoops

deadlybutter: removed

resources/assets/js/split.js
```diff
@@ -0,0 +1,41 @@
+var TOTAL_USERS = $('#competition_max').data('total');
```
weerd: `.data()` for the win 💃

deadlybutter: /giphy dance

weerd:
Cool. Looks good :) Excited to try this out! 👍

deadlybutter:
💥

PR #231, opened 2016-04-19
Updates to fix showing User activity on single user show page.

### LINKED ISSUE #227: Fix call to missing function
image

Function was removed in #224. Missed additional call to it in a different function.

weerd:
#### What's this PR do?
This PR fixes an error with a missing function call that was introduced after #224 was merged. This cleans up and fixes that issue using a similar approach as some of the other methods in the `Manager` class. It also adds a few helpful helper functions.

Load a competition with users, click on a user to view their individual page, and see if their user activity is displayed for the competition with Reportbacks (if any), etc.

Fixes #227

---
@DoSomething/gladiator

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -2,11 +2,12 @@
 
 namespace Gladiator\Http\Controllers;
 
+use Gladiator\Http\Requests\UserRequest;
+use Gladiator\Models\Competition;
 use Gladiator\Models\User;
+use Gladiator\Repositories\UserRepositoryContract;
 use Gladiator\Services\Manager;
 use Gladiator\Services\Registrar;
-use Gladiator\Http\Requests\UserRequest;
-use Gladiator\Repositories\UserRepositoryContract;
 
 class UsersController extends Controller
 {
@@ -97,20 +98,20 @@ class UsersController extends Controller
      */
     public function show($id)
     {
+        $activities = [];
+
         $user = $this->repository->find($id);
 
-        $competitions = User::find($id)->competitions;
+        $competitions = User::findOrFail($id)->competitions;
 
-        // Get the user's reportback for each competition they are in.
         foreach ($competitions as $competition) {
-            $reportback = $this->manager->getActivityForUser($id, $competition);
+            $parameters = $this->manager->getCampaignParameters($competition);
+            $parameters['users'] = $user->id;
 
-            if ($reportback) {
-                $competition->reportback = $reportback;
-            }
+            $activities[] = $this->manager->appendReportback($competition, $parameters);
         }
 
-        return view('users.show', compact('user', 'competitions'));
+        return view('users.show', compact('user', 'activities'));
     }
 
     /**
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -2,12 +2,11 @@
 
 namespace Gladiator\Http\Controllers;
 
-use Gladiator\Http\Requests\UserRequest;
-use Gladiator\Models\Competition;
 use Gladiator\Models\User;
-use Gladiator\Repositories\UserRepositoryContract;
 use Gladiator\Services\Manager;
 use Gladiator\Services\Registrar;
+use Gladiator\Http\Requests\UserRequest;
+use Gladiator\Repositories\UserRepositoryContract;
 
 class UsersController extends Controller
 {
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -73,6 +73,17 @@ function format_date_form_field($model, $field, $defaut = 'MM/DD/YYYY')
 }
 
 /**
+ * Formate a timestamp using Carbon for display in a view.
+ *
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -97,20 +97,20 @@ public function store(UserRequest $request)
...
-        // Get the user's reportback for each competition they are in.
         foreach ($competitions as $competition) {
-            $reportback = $this->manager->getActivityForUser($id, $competition);
+            $parameters = $this->manager->getCampaignParameters($competition);
+            $parameters['users'] = $user->id;
 
-            if ($reportback) {
-                $competition->reportback = $reportback;
-            }
+            $activities[] = $this->manager->appendReportback($competition, $parameters);
```
weerd: Ended up saving the updated competition to the activities array, but could potentially just be reduced to override the competition in the collection? Thoughts?

resources/views/users/show.blade.php
```diff
@@ -1,3 +1,5 @@
+{{-- dd($activities) --}}
```
sbsmith86: should this be removed?

app/helpers.php
```diff
@@ -73,6 +73,17 @@ function format_date_form_field($model, $field, $defaut = 'MM/DD/YYYY')
 }
 
 /**
+ * Formate a timestamp using Carbon for display in a view.
+ *
+ * @param  string  $timestamp
+ * @return /Carbon/Carbon
+ */
+function format_timestamp_for_display($timestamp)
```
sbsmith86: You end up using this function on the `updated_at` field which is already a `Carbon` instance (at least, it should be) and then in this function you make a new Carbon instance out of it again. Is that on purpose?

resources/views/users/show.blade.php
```diff
@@ -38,12 +40,11 @@
                             <th class="table__cell">Items</th>
                             <th class="table__cell">Quantity</th>
                             <th class="table__cell">Updated At</th>
-                            <th class="table__cell">Flagged</th>
```
sbsmith86: we don't need this anymore?

resources/views/users/show.blade.php
```diff
@@ -55,20 +56,20 @@
...
-                                <td class="table__cell">{{ $competition->reportback->flagged or 'N/A' }}</td>
-                                <td class="table__cell"><a href="{{ route('competitions.users.destroy', ['user' => $user->id, 'competition' =>$competition->id]) }}" class="button -danger">Remove</a></td>
+
+                                @if ($competition->reportback)
+                                    <td class="table__cell"><a href="{{ reportback_admin_url($competition->reportback->id) }}">{{ $competition->reportback->id }}</a></td>
+                                    <td class="table__cell">{{ $competition->reportback->reportback_items->total }}</td>
+                                    <td class="table__cell">{{ $competition->reportback->quantity }}</td>
+                                    <td class="table__cell">{{ format_timestamp_for_display($competition->reportback->updated_at) }}</td>
+                                @else
+                                    <td class="table__cell">N/A</td>
```
sbsmith86: Why can't you use the `or` operator here anymore?

resources/views/users/show.blade.php
weerd: doh!

app/helpers.php
weerd: It's not a Carbon instance when retrieved. It's the Reportback data coming from Phoenix, which is just a straight-up regular object.

resources/views/users/show.blade.php
weerd: I didn't think so. Especially since what we actually care about is the statuses of the RB items, not the Reportback parent object.

resources/views/users/show.blade.php
weerd: Not really because it needs to check for nested properties of the reportback object if the `$competition->reportback` object exists. There's no way to easily use `or` if it's not the direct item you're looking for.

So say if I did this:
```php
$competition->reportback->quantity or 'NA'
```

If `$competition->reportback` was `null` to begin with I believe it would fail cause then it's trying to get a `quantity` property on `null`.

resources/views/users/show.blade.php
weerd: But I could defs be wrong! @mikefantini do we _need_ the RB flagged stuff in this table view?

resources/views/users/show.blade.php
weerd: It's also cause now I'm just appending the Reportback to the competition. If there isn't any, it's `null`. Wanted to go about using similar approaches w/ how RBs are appended to collection of users, single user, or competition.

app/helpers.php
sbsmith86: oh ok. i see.

resources/views/users/show.blade.php
sbsmith86: got it.

sbsmith86:
looks good to me 👍

weerd:
thanks @sbsmith86 💃

PR #233, opened 2016-04-21
Removing the pro tip text from the default emails

### LINKED ISSUE #217: If there is no pro tip then "Pro tip - " shouldn't appear
If pro tip field is blank then we should cut the whole pro tip line all together, not have "Pro tip - " and then blank space.

image

deadlybutter:
#### What's this PR do?
What the title says

Does it say pro tip in the email even if there isnt a super pro tip?

See comments in issue below

Fixes #217

### DIFF
--- a/app/Http/Utilities/Correspondence.php
+++ b/app/Http/Utilities/Correspondence.php
@@ -71,7 +71,7 @@ class Correspondence
             'key' => '1',
             'label' => 'Leaderboard update',
             'subject' => '1st :campaign_title: Leaderboard Update',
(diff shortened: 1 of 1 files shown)


angaither:
@deadlybutter let's make sure to communicate this change with Chris/Hilary since we will need to change what the current messaging is saved as already. 👍

deadlybutter:
@angaither should we merge and then update existing stuff, or fix existing stuff then merge?

deadlybutter:
cause i can probably fix this real quick on prod

angaither:
@deadlybutter sure! go ahead merge and then fix

deadlybutter:
updated

PR #235, opened 2016-04-21
Emails with leaderboard

sbsmith86:
## Work in progress

Trying to get some eyes on this so that I can figure out how to properly instantiate the `Manager` class into the `Email` class.

The issue seems to be in `app/Services/Email.php` on [line 50](https://github.com/DoSomething/gladiator/compare/emails-with-leaderboard?expand=1#diff-4f39f258033e5d77d4d966ae87da96c2R50). I thought I could just type-hint a `Manager $manager` in the constructor but i get errors.

The `Email` class is used [here](https://github.com/DoSomething/gladiator/blob/master/app/Listeners/QueueMessage.php#L36)

### DIFF
--- a/app/Listeners/QueueMessage.php
+++ b/app/Listeners/QueueMessage.php
@@ -6,7 +6,7 @@ use Illuminate\Mail\Mailer;
 use Gladiator\Events\QueueMessageRequest;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Gladiator\Models\Message;
-use Gladiator\Http\Utilities\Email;
+use Gladiator\Services\Email;
 
 class QueueMessage implements ShouldQueue
 {
@@ -33,7 +33,7 @@ class QueueMessage implements ShouldQueue
         $resources = $event->resources;
 
         // Build the email.
-        $email = new Email($resources['message'], $resources['contest'], $resources['competition'], $resources['users']);
+        $email = new Email($resources);
 
         foreach ($email->allMessages as $content) {
             $settings = [
--- a/app/Listeners/QueueMessage.php
+++ b/app/Listeners/QueueMessage.php
@@ -33,7 +33,8 @@ class QueueMessage implements ShouldQueue
         $resources = $event->resources;
 
         // Build the email.
-        $email = new Email($resources);
+        $manager = app(\Gladiator\Services\Manager::class);
+        $email = new Email($resources, $manager);
 
         foreach ($email->allMessages as $content) {
             $settings = [
--- a/app/Listeners/QueueMessage.php
+++ b/app/Listeners/QueueMessage.php
@@ -35,18 +35,16 @@ class QueueMessage implements ShouldQueue
         // Build the email.
         $email = new Email($resources['message'], $resources['contest'], $resources['competition'], $resources['users']);
 
-        foreach ($email->allMessages as $message) {
-            $content = $message['message'];
-
+        foreach ($email->allMessages as $content) {
             $settings = [
-                'subject' => $content->subject,
+                'subject' => $content['message']['subject'],
                 'from' => $email->contest->sender_email,
                 'from_name' => $email->contest->sender_name,
-                'to' => $message['user']->email,
-                'to_name' => $message['user']->first_name,
+                'to' => $content['user']->email,
+                'to_name' => $content['user']->first_name,
             ];
 
-            $this->sendMail($content, $settings);
+            $this->sendMail($content['message'], $settings);
         }
     }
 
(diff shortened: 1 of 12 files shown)


app/Services/Email.php
```diff
@@ -46,12 +47,15 @@ class Email
...
-        $this->contest = $contest;
-        $this->competition = $competition;
-        $this->users = $users;
+        $this->message = $resources['message'];
+        $this->contest = $resources['contest'];
+        $this->competition = $resources['competition'];
+        $this->users = $resources['users'];
+        // @TODO - is there a better way of instantiating and pulling in the manager class?
+        // Maybe this email class needs to be a service instead.
+        // $this->manager = $manager;
```
weerd: Where you adding it to the construct?

weerd: to do `$this->manager = $manager`

you need this:

```
public function __construct(Manager $manager, $resources) {
  ...
  $this->manager = $manager;
}
```

app/Listeners/QueueMessage.php
```diff
@@ -33,20 +33,18 @@ public function handle(QueueMessageRequest $event)
         $resources = $event->resources;
 
         // Build the email.
-        $email = new Email($resources['message'], $resources['contest'], $resources['competition'], $resources['users']);
-
-        foreach ($email->allMessages as $message) {
-            $content = $message['message'];
+        $email = new Email($resources);
```
DFurnes: Yeah, so I think this is what's causing you trouble! You can only do dependency injection "magically" in the constructor if Laravel is creating the class for you (through injection in another class or the `app()` helper or something like that). So when you add `Manager $manager` to the constructor on line 50 in `Email.php` it expects that parameter to be passed here.

weerd: Yep! And example in my comment below to adding it to construct via Laravel's smartness vs you having to pass it in here.

app/Services/Email.php
sbsmith86: @weerd m not yet, sorry i was just putting this up there for eyes, but what i was trying to do was 

```
function __contruct($resources, Manager $manager) { }
```

but then it is not injecting the class magically, it expects the second param to be a Manager instance when I try to instantiate the `Email` class

DFurnes: I think @sbsmith86 commented this out after trying that and getting errors. The quickest fix would probably be to change it to `$this->manager = app(Gladiator\Services\Manager::class)` to ask for it explicitly.

DFurnes: I think longer term refactoring it's probably a good idea to try to keep a boundary between "helper objects" that contain values and get manually instantiated and passed around, versus more "global" services that ask for other relevant services from the IoC container.

weerd: oh gotcha! sorry.

app/Listeners/QueueMessage.php
sbsmith86: @weerd @DFurnes yea but then how do i instantiate the `Manger` class before I pass it to the `Email` class? Can I do magical injection in the Listener construct?

app/Services/Email.php
weerd: Yeah... I think part of the issue is that `Email` got _upgraded_ from a helper object to a service.

app/Listeners/QueueMessage.php
DFurnes: I think I'd recommend using the `app()` helper to specifically request the class from the service container inside the `Email` class so you don't have to remember to manually inject and pass along anywhere you use Email elsewhere.

weerd: @DFurnes can probs just use

```
$this->manager = app('manager');
```
right?

app/Services/Email.php
DFurnes: Yeah totes, definitely makes sense as something to gradually refactor. 😄

sbsmith86: yea now that we are doing Leaderboard and Reportback stuff, the `Email` class actually does need to talk to different services. That is why i moved it here, but i can move it back if it doesn't make sense.

app/Listeners/QueueMessage.php
DFurnes: If you've manually created a named binding for the manager in your AppServiceProvider, you could do that. I kinda like using the full class path since it's more explicit about what is being created & doesn't require any extra setup.

app/Services/Email.php
weerd: Yeah, totally makes sense @sbsmith86 :) Just adds a bit of that unavoidable confusion as it gets readjusted!

app/Listeners/QueueMessage.php
weerd: Gotcha. Good to know. Manager doesn't have a dedicated binding, so the FUUUUURNS approach is the one to go with 💃

sbsmith86: Thank you all, you beautiful minds...i will try this out .

PR #236, opened 2016-04-21
Moves contestants to its own index page

### LINKED ISSUE #234: Update the users page
We should remove contestants section
it takes too long to load & I don't think it very useful

if we _really_ need a list of everyone, maybe this should be it's own view.

deadlybutter:
#### What's this PR do?
Moves contestants into there own users index page.

Does the contestants page display everyone?
Does the normal users page still work?

1- Eventually we can use pagination and alleviate the problem entirely 
2- The route was intentionally placed above the resource. If it was below the resource I got a 404 returned.

Fixes #234


app/Http/routes.php
```diff
@@ -42,6 +42,7 @@
     Route::get('messages/{message}/send', 'MessagesController@sendMessage')->name('messages.send');
 
     // Users
+    Route::get('users/contestants', 'UsersController@contestantIndex')->name('users.contestants');
```
deadlybutter: before someone says something about this being above the resource, see what I wrote in #context

angaither: was it 404-ing because of the `Route::get('users/{id}/signup/{signup_id}'` below?

weerd: Yeah that's fine. You get a 404 cause it was trying to look up a user via `/users/:id` and using `contestant` as the id.

deadlybutter: so I can leave it here?

angaither: you should be able to put it between the first and second

weerd: Ya. There's nothing else that matches it, so if the second item in a url param to `/users/?` doesn't match `contestant` then it moves along to next route which is the resource.

deadlybutter: I had it in the middle, thats where it was 404'ing

app/Http/Controllers/UsersController.php
```diff
@@ -158,4 +157,16 @@ public function showSignup($id, $signup_id)
     {
         var_dump('show signup');
     }
+
+    /**
+     * Returns an index of all contestants.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function contestantIndex()
```
weerd: plural maybe? `contestantsIndex`?

deadlybutter: yeah that makes sense

deadlybutter: done

resources/views/users/contestants.blade.php
```diff
@@ -0,0 +1,13 @@
...
+
+    @include('layouts.header',[
+        'title' => 'Contestants'
+    ])
+
+    @if ($contestants->count())
+        @include('users.partials._table_users', ['users' => $contestants, 'role' => 'Contestants'])
+    @endif
+
+@stop
```
weerd: Maybe this file should be `/users/contestants/index.blade.php`. Since we might have a `/users/contestants/show.blade.php`? Or maybe just getting too into the weeds? Otherwise have it be called `contestants_index.blade.php` so it's more obvious the index view, since it's the pattern we've used in similar naming.

deadlybutter:
@weerd renamed

app/Http/Controllers/UsersController.php
```diff
@@ -158,4 +157,16 @@ public function showSignup($id, $signup_id)
     {
         var_dump('show signup');
```
weerd: ?

weerd: haha. Has this been in here the whole time?

weerd:
👍  have at it!

Issue #229, opened 2016-04-19
Pull in the curated reportback


angaither:
created text field for rbid and reportback item id to pull in the curated reportback

sbsmith86:
PR #245 has been merged so closing this.

Issue #206, opened 2016-04-13
Old contests don't have all the emails set up


mikefantini:
https://qa-gladiator.dosomething.org/contests/27/messages/edit is missing "Leaderboard update 2 email"

sbsmith86:
I am seeing this on dev as well. From what I can gather, some contests have messages set in the DB from like a 1st iteration of messages which are the same ones we are using now. 

Older contests have these messages set:

**labels**
`Leaderboard Update`
`Final Leaderboard Update`
`Reminder for first submission due`
`Reminder to submit v1`
`Reminder to submit v2`
`Last minute reminder v1`
`Last minute reminder v2`
`Welcome`

However in the `Correspondence` class where defaults are defined we have (assuming these got put in place after we added the messages above to the older contests)

`Welcome`
`Competition info`
`Where are you? Check status`
`Reminder for first submission due`
`Reminder for submission photo due`
`Almost over reminder`
`Last minute reminder`
`Leaderboard update`
`Leaderboard update 2`
`Final leaderboard update`

**Possible solutions**
On QA, I think we could just wipe and re-seed the DB so that it sets each contest with the newer messages, but there could be an impact on anything we have set up there already for testing purposes.

If this is a problem on Prod, we need a better solution...maybe a seeder that gets run once to update those old contests with the correct default message?

@weerd thoughts?

weerd:
Yeah, I would say wipe and reseed QA, especially since now we have the capacity to also link _real_ staging campaigns instead of a bunch of fake numbers.  On Prod though, not sure. Might need to chat IRL... would need to potentially add an artisan command to repopulate those when missing?

weerd:
I'm thinking a script that fills in missing stuff on Prod is i

weerd:
sorry. pressed wrong button 😬

weerd:
Damn it! Did it again!

sbsmith86:
i'm down to chat about how to handle this IRL @weerd

sbsmith86:
Chatted with @weerd I'm gonna add a command we can run to retroactively update default messages for old contests

So we would just run `php artisan updateDefaultMessages -override` to reset all messages to the defaults set in the `Correspondence` class. the `-override` option will allow us to specify even that for messages that are already in the DB, they need to be reset to the default. Running the command without the `-override` option will just add any new default messages that are missing without overriding any changes to other messages in the system.

weerd:
👍

weerd:
[...]

Issue #195, opened 2016-04-11
Pull reportbacks and the leaderboard into email messages


sbsmith86:
Add top three reportbacks and leaderboard to leaderboard email messages

**Reportbacks**
Each reportback needs to include: first_name, caption, reportback_noun, reportback_quantity, reportback_photo

**Leaderboard**
Should include the entire leaderboard minus people who haven't reported back or who have been flagged.

PR #237, opened 2016-04-21
Emails with leaderboard

sbsmith86:
#### What's this PR do?
This PR is part 1 of getting leaderboards in Email messages. 

* Create a new `signoff` field that admins can use to place text after the leaderboard is injected into the email template.

* Updates how we were creating process messages, before I was making a processed version of the `Message` instance for each user, when really we just needed a few of the processed fields. So now it just creates an array with the fields we need. See below for way an array is needed and fine for this.

* Passing an `Array` containing all the variables needed for output to the message view instead of an `Object` because when using `Mail::queue` Laravel will create a raw serialized array of what is passed to the view, so we need to use array notation anyway. 

* Updates the message views accordingly.

Pushing this up now for review and so we can unblock #229 and #206 

Addresses #195

### DIFF
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -9,7 +9,6 @@ use Gladiator\Models\Message;
 use Gladiator\Events\QueueMessageRequest;
 use Gladiator\Repositories\MessageRepository;
 use Gladiator\Repositories\UserRepositoryContract;
-use Gladiator\Http\Utilities\Email;
 use Gladiator\Services\Manager;
 
 class MessagesController extends Controller
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -104,7 +104,7 @@ class MessagesController extends Controller
                 'email' => $contest->sender_email,
             ];
         } else {
-            $users = $this->manager->getModelUsers($competition);
+            $users = $this->manager->getModelUsers($competition, true);
         }
 
         $resources = [
--- a/app/Listeners/QueueMessage.php
+++ b/app/Listeners/QueueMessage.php
@@ -6,7 +6,7 @@ use Illuminate\Mail\Mailer;
 use Gladiator\Events\QueueMessageRequest;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Gladiator\Models\Message;
-use Gladiator\Http\Utilities\Email;
+use Gladiator\Services\Email;
 
 class QueueMessage implements ShouldQueue
 {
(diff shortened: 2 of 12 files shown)


app/Http/Controllers/MessagesController.php
```diff
@@ -104,7 +103,7 @@ public function sendMessage(Message $message)
                 'email' => $contest->sender_email,
             ];
         } else {
-            $users = $this->manager->getModelUsers($competition);
+            $users = $this->manager->getModelUsers($competition, true);
```
weerd: Yay! Reportbacks 💃

app/Listeners/QueueMessage.php
```diff
@@ -33,20 +33,19 @@ public function handle(QueueMessageRequest $event)
         $resources = $event->resources;
 
         // Build the email.
-        $email = new Email($resources['message'], $resources['contest'], $resources['competition'], $resources['users']);
-
-        foreach ($email->allMessages as $message) {
-            $content = $message['message'];
+        $manager = app(\Gladiator\Services\Manager::class);
+        $email = new Email($resources, $manager);
```
weerd: So might be easier than always needing to remember to pass in an instance of the `Manager` class when creating an instance of `Email` to add it via the constructor of `Email` but not as an argument:

```php
// Email class

protected $manager;

public function __construct($resources)
{
  $this->manager = app(\Gladiator\Services\Manager::class);
}
```

If it needs it, now it's always included 😉

sbsmith86: ahhh yep that is what y'all said. haha! makes sense.

weerd: 💃

sbsmith86:
@weerd made that update!

weerd:
We should rename this PR to `REDRUM` in honor of room #237. My parent's house on long island is #237 and I got super freaked out when I was younger when I found out that was the hotel number of the all the creepy shit in the shining.

Sorry, random 💭

sbsmith86:
@weerd that sounds like a horror movie reference and I do not subscribe to those. 😉

weerd:
Haha, totally. Neither do I until it was too late to realize what I was watching was a horror movie. 😱  🔪

app/Services/Email.php
```diff
@@ -137,6 +139,16 @@ protected function setupEmail()
...
+
+            // Leaderboard messages get an extra leaderboard variable to be sent to the email template.
+            // @TODO - move into smaller function that gets the leaderboard and then add it to the allMessages array here.
+            // if ($this->message->type == 'leaderboard') {
+            //     $list = $this->manager->catalogUsers($this->users);
+
+            //     $this->allMessages[$key]['message']['leaderboard'] = $list['active'];
+            // }
+
+            // @TODO - create a smaller function that can be called here that gets the top 3 reportbacks.
```
weerd: Comments for future reference right? For the upcoming leaderboard addition to messages?

sbsmith86:
@weerd haha sooooo ima merge this then?

weerd:
I think this looks good to me 👍

app/Services/Email.php
sbsmith86: @weerd yes!

PR #238, opened 2016-04-22
Quick update to allow easily adding mutators for Message attributes.

weerd:
#### What's this PR do?
This PR does a quick update to how Message updating is handled and allows setting up mutators in the Message model for the different attributes we are adding as messages evolves.

Try editing messages for a contest, particularly the `pro_tip` and `signoff` fields and make sure the values you entered stick after submitting the form.

@angaither looked further into this, and it seems the capability to use the `set*Attribute()` method on the model depends on right before saving the model, that the actual attribute is directly called. So if you want to create a setter for `pro_tip`, you can't just use the `fill()` method and pass it the input, you specifically need to do:

```php
$message->pro_tip = $data['pro_tip']
```

The above actually activates the `setProTipAttribute($value)` method on the `Message` model.

You can find more info here: https://laravel.com/docs/5.2/eloquent-mutators#accessors-and-mutators

I did take your advice though to make it easier when setting the attributes by first grabbing the array items from the `fillable` property, and [did it in a way](https://github.com/DoSomething/gladiator/compare/messages-update-refactor?expand=1#diff-0611f5ce25165dc163f30daf92ee8c7bR58) that still allows accessing `setter` mutators on each attribute 💃 

Refs #206

### DIFF
--- a/app/Repositories/MessageRepository.php
+++ b/app/Repositories/MessageRepository.php
@@ -16,7 +16,6 @@ class MessageRepository
     public function create($contest, $data)
     {
         $message = new Message($data);
-        $message->pro_tip = empty($data['pro_tip']) ? null : $data['pro_tip'];
 
         return $contest->messages()->save($message);
     }
@@ -51,9 +50,14 @@ class MessageRepository
     public function update($contest, $data)
     {
         $message = Message::where('contest_id', '=', $contest->id)->where('type', '=', $data['type'])->where('key', '=', $data['key'])->firstOrFail();
-        $message->subject = $data['subject'];
-        $message->body = $data['body'];
-        $message->pro_tip = empty($data['pro_tip']) ? null : $data['pro_tip'];
+
+        $attributes = $message->getFillable();
+
+        foreach ($attributes as $attribute) {
+            if (isset($data[$attribute])) {
+                $message->{$attribute} = $data[$attribute];
+            }
+        }
 
         return $message->save();
     }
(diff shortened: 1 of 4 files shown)


app/Repositories/MessageRepository.php
```diff
@@ -16,7 +16,6 @@ class MessageRepository
     public function create($contest, $data)
     {
         $message = new Message($data);
-        $message->pro_tip = empty($data['pro_tip']) ? null : $data['pro_tip'];
```
weerd: Just to note, we only really run `create` for Messages during the seeding process and that grabs from the defaults in the `Correspondence` class which is already sets the `pro_tip` to `null`. So no real need to bother with the mutator... at least until we add a `Create Messages` form, which I don't is planned to happen at any point soon 😉

app/Repositories/MessageRepository.php
```diff
@@ -51,9 +50,14 @@ public function createMessagesForContest($contest, $messages)
...
         $message = Message::where('contest_id', '=', $contest->id)->where('type', '=', $data['type'])->where('key', '=', $data['key'])->firstOrFail();
-        $message->subject = $data['subject'];
-        $message->body = $data['body'];
-        $message->pro_tip = empty($data['pro_tip']) ? null : $data['pro_tip'];
+
+        $attributes = $message->getFillable();
+
+        foreach ($attributes as $attribute) {
+            if (isset($data[$attribute])) {
+                $message->{$attribute} = $data[$attribute];
```
angaither: ahh much nicer! 👏

angaither:
👍

weerd:
💃

PR #239, opened 2016-04-22
Emails with leaderboard 2

sbsmith86:
#### What's this PR do?
Pulls in the leaderboard to leaderboard messages. Uses a partial to layout the message. Very basic styling at the moment, HTML email styling is nightmare and we should probably think of how the email should look as a whole.

screen shot 2016-04-21 at 11 56 50 pm

This just deals with the leaderboard, next I will deal with pulling in the top three reportbacks. 

I also had to update how the test emails work, now they send just the first message that is processed to that admin. I put a `TODO` to fix this. If this is a test email, we should only process one message and send that to the admin. 

Addresses #195


weerd:
@sbsmith86 very tiny conflict. I think it's cause I changed the comment in the * *_form_contest_messaging.blad.php** file to use blade commenting in a PR I just got approved and merged. Should be a tiny change 😬

sbsmith86:
@weerd fixed.

angaither:
👍

PR #244, opened 2016-04-22
leaderboards with top three rbs

### LINKED ISSUE #195: Pull reportbacks and the leaderboard into email messages

sbsmith86:
#### What's this PR do?
Pulls in the top three reportbacks to leaderboard emails. 

Desktop
image

general idea of what this looks like on phones:

image

Pull down, send a leaderboard email on a competition with a leaderboard, you should get that email with the top three reportbacks in it. 

Fixes #195

### DIFF
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -133,6 +133,8 @@ class Email
      */
     protected function setupEmail()
     {
+        $leaderboardVars = $this->getLeaderboardVars($this->users);
+
         // Each user gets it's own processed message
         foreach ($this->users as $key => $user) {
             $this->allMessages[$key]['user'] = $user;
@@ -143,8 +145,8 @@ class Email
 
             $message = $this->processMessage($tokens, $this->message);
 
-            if ($this->message->type == 'leaderboard') {
-                $message = array_merge($message, $this->getLeaderboardVars($this->users));
+            if ($this->message->type === 'leaderboard') {
+                $message = array_merge($message, $leaderboardVars);
             }
 
             $this->allMessages[$key]['message'] = $message;
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -165,6 +165,7 @@ class Email
         $vars = [
             'leaderboard' => $leaderboard,
             'topThree' => $this->manager->getTopThreeReportbacks($leaderboard),
+            'reportbackInfo' => $this->contest->campaign->reportback_info,
         ];
 
         return $vars;
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -139,11 +139,15 @@ class Email
 
             $tokens = $this->defineTokens($user);
 
-            $this->allMessages[$key]['message'] = $this->processMessage($tokens, $this->message);
+            $message = [];
+
+            $message = $this->processMessage($tokens, $this->message);
 
             if ($this->message->type == 'leaderboard') {
-                $this->allMessages[$key]['message'] = array_merge($this->allMessages[$key]['message'], $this->processLeaderboardVars($this->users));
+                $message = array_merge($message, $this->processLeaderboardVars($this->users));
             }
+
+            $this->allMessages[$key]['message'] = $message;
         }
     }
 
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -129,7 +129,7 @@ class Email
     }
 
     /**
-     * Builds the email object.
+     * Builds the email array.
      */
     protected function setupEmail()
     {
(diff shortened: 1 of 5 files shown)


sbsmith86:
@angaither @weerd @deadlybutter Anyone mind giving this a 👀

deadlybutter:
looks 👍  to me

app/Services/Email.php
```diff
@@ -138,25 +138,36 @@ protected function setupEmail()
...
 
-            if ($this->message->type == 'leaderboard') {
-                $this->allMessages[$key]['message']['leaderboard'] = $this->generateLeaderboard($this->users);
+            $message = [];
+
+            $message = $this->processMessage($tokens, $this->message);
 
-                // @TODO - get top 3 reportbacks
+            if ($this->message->type == 'leaderboard') {
+                $message = array_merge($message, $this->getLeaderboardVars($this->users));
```
angaither: does this need to be processed within the `foreach`? it's the same for every user, right?

sbsmith86: @angaither ah yep. good point.

app/Services/Email.php
```diff
@@ -138,25 +138,36 @@ protected function setupEmail()
...
-            $this->allMessages[$key]['message'] = $this->processMessage($tokens, $this->message);
 
-            if ($this->message->type == 'leaderboard') {
-                $this->allMessages[$key]['message']['leaderboard'] = $this->generateLeaderboard($this->users);
+            $message = [];
+
+            $message = $this->processMessage($tokens, $this->message);
 
-                // @TODO - get top 3 reportbacks
+            if ($this->message->type == 'leaderboard') {
```
weerd: triple equals! Aiming to keep us all using triple equals as much as possible. Keeps us honest 😉

app/Services/Manager.php
```diff
@@ -254,6 +254,32 @@ public function appendReportback($data, $parameters)
...
+     * @param array $leaderboard
+     * @return array $topThree
+     */
+    public function getTopThreeReportbacks($leaderboard)
+    {
+        $topThreeUsers = array_slice($leaderboard, 0, 3);
+        $places = ['1st', '2nd', '3rd'];
+
+        foreach ($topThreeUsers as $key => $user) {
+            $latestReportback = array_pop($user->reportback->reportback_items->data);
```
weerd: Super nit-picky, but could we call this variable `$latestReportbackItem` to really drive the point home it's the item and not the general reportback? Try to make sure the terminology is applied across projects as much as possible cause there's already enough confusion between Reportbacks and Reportback Items @_@

sbsmith86: makes sense! will do.

app/Services/Manager.php
```diff
@@ -254,6 +254,32 @@ public function appendReportback($data, $parameters)
...
+            $topThree[] = [
+                'place' => $places[$key],
+                'first_name' => $user->first_name,
+                'quantity' => $user->reportback->quantity,
+                'image_url' => $latestReportback->media->uri,
+                'caption' => $latestReportback->caption,
+            ];
+        }
+
+        return $topThree;
```
weerd: Just curious... no errors/warnings thrown for not defining the `$topThree` variable outside the `foreach`?

weerd:
Just a few comments, but looks awesome otherwise 💃

app/Services/Manager.php
sbsmith86: @weerd yea no errors. It's a php 5 thing. You can use square bracket syntax to create, modify arrays. 

> If $arr doesn't exist yet, it will be created, so this is also an alternative way to create an array.

sbsmith86:
@angaither @weerd addressed your comments

angaither:
🚢

app/Services/Manager.php
```diff
@@ -254,6 +254,33 @@ public function appendReportback($data, $parameters)
...
+            $topThree[] = [
+                'place' => $places[$key],
+                'first_name' => $user->first_name,
+                'quantity' => $user->reportback->quantity,
+                'image_url' => $latestReportbackItem->media->uri,
+                'caption' => $latestReportbackItem->caption,
+            ];
+        }
+
+        return $topThree;
```
weerd: I think if you don't assign `$topThree = [];` up top, outside the foreach loop, then it very well could fail, if the `$leaderboard` is an empty array and the foreach never runs and never creates the `$topThree` variable? _I think_ 🤔

PR #245, opened 2016-04-22
Featured reportback (WIP)

angaither:
#### What's this PR do?
- adds featured reportback item to message leaderboard forms
- created function to grab one item from the reportbacks endpoint

Pull down, add values to the reportback id and reportback item id

I tried using the `reportback-items` endpoint, but it doesn't seem to work see DoSomething/phoenix/issues/6377

Also this does _not_ yet add the reportback item to the email. waiting for code from #244

Refs #229

### DIFF
--- a/app/Models/Message.php
+++ b/app/Models/Message.php
@@ -11,7 +11,7 @@ class Message extends Model
      *
      * @return array
      */
-    protected $fillable = ['contest_id', 'type', 'key', 'label', 'subject', 'body', 'pro_tip', 'signoff', 'reportback_id', 'reportback_item_id'];
+    protected $fillable = ['contest_id', 'type', 'key', 'label', 'subject', 'body', 'pro_tip', 'signoff', 'reportback_id', 'reportback_item_id', 'shoutout'];
 
     /**
      * Excluded attributes that are not customizeable.
--- a/app/Models/Message.php
+++ b/app/Models/Message.php
@@ -11,7 +11,7 @@ class Message extends Model
      *
      * @return array
      */
-    protected $fillable = ['contest_id', 'type', 'key', 'label', 'subject', 'body', 'pro_tip', 'signoff'];
+    protected $fillable = ['contest_id', 'type', 'key', 'label', 'subject', 'body', 'pro_tip', 'signoff', 'reportback_id', 'reportback_item_id'];
 
     /**
      * Excluded attributes that are not customizeable.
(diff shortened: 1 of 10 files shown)


app/Models/Message.php
```diff
@@ -56,6 +56,26 @@ public function setProTipAttribute($value)
     }
 
     /**
+     * Set the reportback_id attribute for the message.
+     *
+     * @param string  $value
+     */
+    public function setReportbackIdAttribute($value)
+    {
+        $this->attributes['reportback_id'] = ! empty($value) ? $value : null;
```
weerd: Maybe switch these to match the `setProTipAttribute()` method above so it uses the same syntax, checking for empty, instead of `!` empty:

```php
public function setProTipAttribute($value)
    {
        $this->attributes['pro_tip'] = empty($value) ? null : $value;
    }
```

Or switch the syntax of `setProTipAttribute()`; whatever is the pref, just make consistent :)

app/Services/Phoenix/Phoenix.php
```diff
@@ -48,4 +48,22 @@ public function getCampaign($id)
...
     }
+
+    /**
+     * Send a GET request to return a reportback & an item with the specified id.
+     *
+     * @param  string $reportback_id
+     * @param  string $reportback_item_id
+     * @return object|null
+     */
+    public function getSpecificReportbackItem($reportback_id, $reportback_item_id)
```
weerd: the pattern we've been using based on original Northstar pattern is to just use `getReportbackItem($id)` to imply retrieving a single resource and `getAllReportbackItems($params = [])` would be for a collections of resources.

app/Services/Phoenix/Phoenix.php
```diff
@@ -48,4 +48,22 @@ public function getCampaign($id)
...
+    {
+        $response = $this->get($this->base_uri . 'reportbacks/' . $reportback_id);
```
weerd: I don't think we need to pass `$this->base_uri .` here. This gets set in the `__construct` for the `Phoenix` class... but it did highlight that in the `getCampaigns()` we're doing that too. We should be able to remove in both and simplify the call to just:

```php
$response = $this->get('reportbacks/' . $reportback_id);
```

angaither:
👍

sbsmith86:
@angaither made a few more commits on this...adding in a `shoutout` field that admins can use to callout the featured reportback. If that is all good, we can merge this to get testing started. But I have a refactor branch I am also working on to clean up some of this.

angaither:
👍

PR #246, opened 2016-04-22
Message refills

### LINKED ISSUE #206: Old contests don't have all the emails set up

weerd:
#### What's this PR do?
This PR updates the `MessageTableSeeder` to work a little smarter, and keep us from having to modify multiple places if/when we add new fields to the default message structure. It then uses a similar approach to build out the new `updated:messages` artisan console command, which will update all messages in a contest(s), and if any are missing, based on the list of defaults, will create the missing ones.

This command can be called in one of two ways:

1) To update **all** messages for **every** contest.
```shell
$ php artisan update:messages
```

2) To update only a specified contest.
```shell
$ php artisan update:messages 12
```
Where `12` is the ID of a specific contest.

This PR also creates a new `buildMessagesFromDefaults()` method on the `MessageRepository` class which builds out all the default messages into an easy to use array.

Head into SequelPro and wreck all kinds of havoc inside the `messages` table. Delete rows, change the values to random strings (I recommend the use of the string `poopie`). Then run the command, refresh SequelPro and see if everything is fixed!

Fixes #206

---
@DoSomething/gladiator

### DIFF
--- a/app/Console/Commands/UpdateMessages.php
+++ b/app/Console/Commands/UpdateMessages.php
@@ -15,7 +15,7 @@ class UpdateMessages extends Command
      *
      * @var string
      */
-    protected $signature = 'update:messages {contest?} {--override : Override all contest messages and refresh with latest defaults }';
+    protected $signature = 'update:messages {contest?}';
 
     /**
      * The console command description.
--- a/app/Console/Commands/UpdateMessages.php
+++ b/app/Console/Commands/UpdateMessages.php
@@ -2,6 +2,7 @@
 
 namespace Gladiator\Console\Commands;
 
+use Gladiator\Models\Contest;
 use Illuminate\Console\Command;
 
 class UpdateMessages extends Command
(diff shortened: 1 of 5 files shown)


app/Console/Commands/UpdateMessages.php
```diff
@@ -0,0 +1,67 @@
...
+
+        if ($contest) {
+            $contest = Contest::findOrFail($contest);
+
+            $this->repository->updateMessagesForContest($contest, $messages);
+
+            return $this->comment(PHP_EOL . 'All set! Messages for Contest ID #' . $contest->id . ' have been updated. If any messages were missing, they were added as well!' . PHP_EOL);
+        }
+
+        $contests = Contest::all();
```
angaither: the first time I read this I thought that all contests would be updated even if you passed in an id. (I missed the return)

maybe have this section be in an else?

angaither:
just one question, but otherwise 👍

app/Console/Commands/UpdateMessages.php
weerd: Kinda going with the keeping code as clean as possible that we saw at that Laracon 2015 😉  Emphasizes only really `else` when it's really needed. But if it's confusing I can change.

PR #247, opened 2016-04-22
No more silly homepage and other cleanup.

weerd:
#### What's this PR do?
This PR fixes the `Contestants` view button to make it a more subtle link... too many of those big buttons up top is jarring on the 👀 : 

image

Also updates the root (home) route so it redirects to the ~~users~~ contests index page instead of the super awesome, **All Your Base Are Belong To Us** welcome page. Sad to see it go, but alas. If you're not logged in, then it'll just send you straight to the login page ⛔ 

Try going to the home page when logged out. Do you see the login form?

Fixes 🌎 

---
@DoSomething/gladiator

### DIFF
--- a/resources/views/pages/dashboard.blade.php
+++ b/resources/views/pages/dashboard.blade.php
@@ -1,0 +1,1 @@
+{{-- More to come. --}}
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -14,7 +14,7 @@
 Route::group(['middleware' => ['web']], function () {
 
     Route::get('/', function () {
-        return redirect()->route('users.index');
+        return redirect()->route('contests.index');
     });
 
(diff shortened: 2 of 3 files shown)


angaither:
😢 🚢

resources/views/pages/dashboard.blade.php
```diff
@@ -0,0 +1 @@
+{{-- More to come. --}}
```
deadlybutter: 👀

deadlybutter:
👍

weerd:
I didn't delete the `home.blade.php` file w/ the awesome homepage... so not all is lost!

deadlybutter:
can we make it an easter egg

weerd:
yes we can

PR #249, opened 2016-04-25
Rest api client updates

### LINKED ISSUE #180: Update RestApiClient for BOTH Northstar & Phoenix

weerd:
#### What's this PR do?
This PR fixes a bug that was occurring when we would receive an error response from the Phoenix API. The `RestApiClient` class was expecting the format for error responses in the way that Northstar formats them, which is slightly different than how Phoenix does. This fix updates the response object to include the expected `code` property in the `error` object.

Do you still get the "Undefined property: stdClass::$code" error?

Tested this a bit with forcing errors from Phoenix, but definitely something to keep an eye out for if it crops up again!

Fixes #180 

---
@DoSomething/gladiator

### DIFF
--- a/app/Services/RestApiClient.php
+++ b/app/Services/RestApiClient.php
@@ -174,7 +174,6 @@ class RestApiClient
      */
     protected function setMessages($error)
     {
-        // dd('errors');
         // @TODO: may eventually need to handle an array of errors.
         return (new MessageBag)->add($error->code, $error->message);
     }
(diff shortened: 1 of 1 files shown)


sbsmith86:
looks ok to me!

Issue #241, opened 2016-04-22
Testing welcome email


mikefantini:
As an admin, I'd like to be able to send myself a test email of the welcome email users receive in #240.

Ideally the user could send this email from the /contests/[cid] view. Perhaps we can add a button to the messaging section of this view to send a test welcome email. Functionally this should work the same way as test emails do for individual competitions (i.e. a user gets sent to the sender email address).

deadlybutter:
after adding this, lets kill it from the messages page as @sbsmith86 mentioned here https://github.com/DoSomething/gladiator/pull/251#issuecomment-214479290

PR #251, opened 2016-04-25
Emails welcome

### LINKED ISSUE #240: Welcome email automatic trigger
The welcome email should be sent to a user as soon as they enter a contest (so once they say they want to join the competition on Phoenix).

So this will no longer be a broadcast email but instead a transactional one.

deadlybutter:
#### What's this PR do?
Adds email sending upon user import to gladiator / competition join.

Ive been using Postman with the following details,

```
POST /api/v1/users HTTP/1.1
Host: gladiator.app
Content-Type: application/json
X-DS-Gladiator-API-Key: hah nope.
Accept: application/json
Cache-Control: no-cache
{
    "id": "joe@dosomething.org",
    "term": "email",
    "role": "user",
    "campaign_run_id": "1903",
    "campaign_id": "1485"
}
```

You'll also need to make sure that ^ test user isnt in the user waiting rooms pivot table.

I had to make some changes to Email because it assumes that a competition has been created. (For welcome emails only the contest exists)

Fixes #240

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -5,6 +5,7 @@ namespace Gladiator\Http\Controllers\Api;
 use Gladiator\Models\User;
 use Gladiator\Models\Contest;
 use Gladiator\Services\Registrar;
+use Gladiator\Services\Manager;
 use Gladiator\Http\Requests\UserRequest;
 use Gladiator\Http\Transformers\UserTransformer;
 use Gladiator\Events\QueueMessageRequest;
@@ -24,12 +25,18 @@ class UsersController extends ApiController
     protected $transformer;
 
     /**
+     * @var Gladiator\Services\Manager;
+     */
+    protected $manager;
+
+    /**
      * Create new UsersController instance.
      */
-    public function __construct(Registrar $registrar)
+    public function __construct(Registrar $registrar, Manager $manager)
     {
         $this->registrar = $registrar;
         $this->transformer = new UserTransformer;
+        $this->manager = $manager;
 
         $this->middleware('auth.api');
     }
@@ -77,6 +84,7 @@ class UsersController extends ApiController
         }
 
         $contest->waitingRoom->users()->attach($user->id);
+        $this->manager->appendCampaign($contest);
 
         // Fire off welcome Email
         $message = Message::where(['contest_id' => $contest->id, 'type' => 'Welcome'])->first();
@@ -84,7 +92,7 @@ class UsersController extends ApiController
             'message' => $message,
             'contest' => $contest,
             'users' => [$user],
-            'test' => false,
+            'test' => true,
         ];
         event(new QueueMessageRequest($resources));
 
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -92,7 +92,7 @@ class UsersController extends ApiController
             'message' => $message,
             'contest' => $contest,
             'users' => [$user],
-            'test' => true,
+            'test' => false,
         ];
         event(new QueueMessageRequest($resources));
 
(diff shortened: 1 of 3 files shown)


app/Services/Email.php
```diff
@@ -70,10 +70,10 @@ protected function defineTokens($user)
         // @TODO - set defaults?
         $tokens = [
             ':campaign_title:'        => $this->contest->campaign->title,
-            ':end_date:'              => $this->competition->competition_end_date->format('F d, Y'),
+            ':end_date:'              => isset($this->competition) ? $this->competition->competition_end_date->format('F d, Y') : '',
```
deadlybutter: See #context in this PR for why I had to do this

deadlybutter:
🆗  ready for review

app/Http/Controllers/Api/UsersController.php
```diff
@@ -74,6 +83,17 @@ public function store(UserRequest $request)
         }
 
         $contest->waitingRoom->users()->attach($user->id);
+        $this->manager->appendCampaign($contest);
+
+        // Fire off welcome Email
+        $message = Message::where(['contest_id' => $contest->id, 'type' => 'Welcome'])->first();
```
sbsmith86: I *think* `Welcome` should be lowercased...that's how it's stored in the db at least.

deadlybutter: yep just checked, lower case!

app/Services/Email.php
sbsmith86: shouldn't you just check if this is null here? Since you set it as `null` in the constructor?

sbsmith86:
Should we remove the option to send the message on the messages page also? Not sure what the benefit of having it now that the messages will get sent automatically.

deadlybutter:
I think we should keep the test button, but the actual send button could go 👍

sbsmith86:
there is an issue to move the test button to the contest create page #241

deadlybutter:
hm, I feel like we shouldnt kill the functionality on the messages page until ^ issue is resolved. but once it is than 💥

sbsmith86:
sounds good

sbsmith86:
looks good to me otherwise 💯

app/Http/Controllers/Api/UsersController.php
```diff
@@ -74,6 +83,17 @@ public function store(UserRequest $request)
         }
 
         $contest->waitingRoom->users()->attach($user->id);
+        $this->manager->appendCampaign($contest);
+
+        // Fire off welcome Email
+        $message = Message::where(['contest_id' => $contest->id, 'type' => 'welcome'])->first();
```
angaither: an annoying post merge-comment... should this be it's own function?

deadlybutter: I mean it could be, I'm not sure if theres much benefit in it though.
- this is (AFAIK) the only place that needs the welcome email
- its 1 line of code, so its not breaking much up

weerd: I think it'd be best as a separate function, as part of the `Message` class or `Email` class (whichever makes more sense), it could allow sending a Welcome message outside of the context of first signup or whatever. Would also help to keep things cleaner. Best to not do a lot in controller; separation of concerns.

deadlybutter: OH YOU MEAN THE ENTIRE oh

i thought you were just referring too the Message:where line

sbsmith86: I don't think the `Email` class should be responsible for firing off events because the event handler ends up instantiating the Email class to build the email object we need for actual sending. Also wary of it being the `Message` class since this needs to gather stuff from many different models. If anything this should be a function in the `Manger` that can be called in the controller.

sbsmith86:
Actually, I am doing some work on refactoring sending one off emails (like the test email). I can include this in that.

PR #256, opened 2016-04-26
Adds official rules

### LINKED ISSUE #243: Official Rules
We need to include official rules in the "Competition info" email. In order to do this we'll need to:

- [x] add a field to the waiting room split form for the link to the Google Doc
- [x] create a token for this field to be referenced in the email body
- [x] update the copy of the email body to end with [official rules](:official_rules_link:). This should go below the :sender_name: (with a paragraph space in between)

deadlybutter:
#### What's this PR do?
Adds official rules to competitions & split view.

- Does the migration add the column?
- Can you split + add official rules?
- Does the :rules: token work?
- Is the default updated?

Competition w/ Rules column + value
screen shot 2016-04-26 at 1 35 13 pm

Test email w/ Rules token 
screen shot 2016-04-26 at 1 35 50 pm

Rules at bottom of competition info email is now the default
screen shot 2016-04-26 at 1 36 48 pm

Rules added to split form
screen shot 2016-04-26 at 1 37 54 pm

In Email on line 148 I had to move the leaderboard variables out of the functions first line because stuff like top3 doesnt exist for a non leaderboard email, and this was giving errors when I tried to send a non leaderboard email. Error below: 

screen shot 2016-04-26 at 1 27 18 pm

Fixes #243


deadlybutter:
@sbsmith86 @weerd or @angaither anyone wanna give this a look?

database/migrations/2016_04_26_151130_add_rules_to_competition.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('competitions', function (Blueprint $table) {
+            $table->string('rules');
```
angaither: add this `after->('leaderboard_msg_day')`

deadlybutter: what does that do?

app/Services/Email.php
```diff
@@ -133,8 +134,6 @@ protected function parseLinks($string)
      */
     protected function setupEmail()
     {
-        $leaderboardVars = $this->getLeaderboardVars($this->users);
```
angaither: I think this was moved out of the loop on purpose, since the leaderboard is the same for all users

deadlybutter: see #context for why I moved it

database/migrations/2016_04_26_151130_add_rules_to_competition.php
angaither: adds it in before the `created_at` and `udpated_at` cols in the db

deadlybutter: oh cool, okie dokes one sec

app/Http/Requests/SplitRequest.php
```diff
@@ -24,6 +24,7 @@ public function rules()
         return [
             'competition_end_date' => 'required|date|after:today',
             'leaderboard_msg_day' => 'required',
+            'rules' => 'required',
```
angaither: I think we could add this should be a url as well https://laravel.com/docs/5.2/validation#rule-url

app/Services/Email.php
angaither: I saw that, but we don't want to process leaderboard vars for each user

deadlybutter: Diego cleaned up a lot of that processing stuff, it only processes the leaderboard once and then it gets stored in flash() -- I can play with moving it out if you prefer tho

database/migrations/2016_04_26_151130_add_rules_to_competition.php
angaither: this should also probably be nullable

app/Services/Email.php
weerd: For context, yes it gets stored in Flash, but only for the next `request` and then it's gone again. I didn't want to store for too long since it's meant to be data that might change quickly. Buuuut wanted to have it in flash for like if you loaded the leaderboard page, and then wanted to export the data of that page (no sense in doing all the heavy processing again for that same info). @deadlybutter not sure if each email cycle would be considered the same request or each is a unique one :shrug:... so flashed data may not be available. Just something to consider since I hadn't intended it to be used outside of a first and second request cycle.

deadlybutter:
just trying to fix one bug but everything else has been fixed @angaither

deadlybutter:
ok apparently there wasnt a bug.... just another case where inserting a `log` statement makes the code work. dunno if this is PHP, laravel, vagrant, blargh.

anyway @angaither i think this is good to merge now

resources/views/waitingrooms/partials/_form_split.blade.php
```diff
@@ -13,6 +13,11 @@
             <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" }/>
         </div>
 
+        <div class="form-item -padded">
+            <label class="field-label" for="rules">Rules URL:</label>
+            <input type="text" name="rules" id="rules" class="text-field" value="docs.google.com/the-rules" />
```
weerd: Is `docs.google.com/the-rules` an actual link?

weerd: If not then don't make this the value, make it the `placeholder` attribute.

deadlybutter: 👍

app/Services/Email.php
sbsmith86: also i fixed the issue with the `$topThree` var in PR #245. I had to define the variable outside of the forloop that was seeing values inside for it as @weerd suggested [here](https://github.com/DoSomething/gladiator/pull/244#discussion_r60791479) So it shouldn't be a problem anymore once you rebase with master.

deadlybutter: oh ok, i just ended up adding a second if(is leaderboard email) but sounds like rebasing would just be cleaner

app/Services/Email.php
```diff
@@ -79,6 +79,7 @@ protected function defineTokens($user)
             ':reportback_noun:'       => strtolower($this->contest->campaign->reportback_info->noun),
             ':reportback_verb:'       => strtolower($this->contest->campaign->reportback_info->verb),
             ':sender_name:'           => $this->contest->sender_name,
+            ':rules:'                 => ! is_null($this->competition) ? $this->competition->rules : '',
```
sbsmith86: Maybe we should call this token `:rules_url:` or something like that. So admins can be clear if they just want to print out the raw URL in the email or if they want to be able to hyperlink with in text with markdown. Similar to how we do with the `prove_it_link`

deadlybutter:
🆗  I think I got everything covered

app/Http/Requests/SplitRequest.php
```diff
@@ -24,6 +24,7 @@ public function rules()
         return [
             'competition_end_date' => 'required|date|after:today',
             'leaderboard_msg_day' => 'required',
+            'rules' => 'required|url',
```
sbsmith86: You use `:rules_ur:` in the body fields of the default messages but didn't change it here.

deadlybutter: Ah yeah makes sense to do that

sbsmith86: oh wait no this is the Request class haha sorry.

weerd: Doesn't hurt to update it to `rules_url` and same in the form field attributes... Would be more explicit which isn't a bad thing ;)

weerd:
Just that comment on `rules_url` potential update... otherwise 👍

deadlybutter:
All updated, seems to still work when I test w/ Mandrill

deadlybutter:
Can I merge this?

weerd:
👍

PR #259, opened 2016-04-27
Makes sure the email is appended

### LINKED ISSUE #240: Welcome email automatic trigger

deadlybutter:
#### What's this PR do?
Forces the email to be appended to the user obj if the term is email.

Are you sent a welcome email upon joining a competition?

Due to some funkyness that i dont entirely understand the User obj returned from the repository doesnt have an email on it. So I just forced it to be there so the email can use it.

Fixes #240

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -69,6 +69,10 @@ class UsersController extends ApiController
             unset($credentials['term']);
 
             $user = $this->registrar->createUser((object) $credentials);
+
+            if ($request->input('term')) {
+                $user->email = $request->input('id');
+            }
         }
 
         $contest = Contest::with(['waitingRoom', 'competitions'])->where('campaign_id', '=', $request['campaign_id'])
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/UsersController.php
```diff
@@ -69,6 +69,10 @@ public function store(UserRequest $request)
             unset($credentials['term']);
 
             $user = $this->registrar->createUser((object) $credentials);
+
+            if ($request->input('term')) {
```
angaither: I would add && `term = email` because I think you can pass multiple types of things in to get a user here

deadlybutter: ah true, could be ID, phone, forgot about that

deadlybutter: fixed @angaither !

deadlybutter: no its not fixed i made a typo, this is why we test

deadlybutter:
ok all tested, works @angaither

angaither:
👍

PR #262, opened 2016-04-27
Created functions to get or remove flagged users from the list.

### LINKED ISSUE #255: Flag flagged users from leaderboards
update leaderboard display to prominently display when people are flagged. 

here's an example
image

### LINKED ISSUE #276: Leaderboard in email and top 3 images pulling in flagged reportbacks
If a reportback is flagged it should not be included in the leaderboard and any images associated with it should not show up in the featured rb images (mentioned in https://github.com/DoSomething/gladiator/pull/262#issuecomment-215438894 as well).

For example, RBID 4059 (screenshot of Jing) is flagged but is still showing up for users as the top reportback and the user who submitted it (Max) is showing up at the top of the leaderboard.

image

angaither:
#### What's this PR do?
removes users with a flagged reportback out of the leaderboard display and leaderboard emails

take a look at a leaderboard with flagged users, they should no longer appear here `/competitions/X/leaderboard`

I left some functions in there that are unused, because they are 🆒 

fixes #255 fixes #276

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -172,10 +172,11 @@ class CompetitionsController extends Controller
         $list = $this->manager->catalogUsers($users);
 
         $leaderboard = $list['active'];
+        $flagged = $list['flagged'];
         $pending = $list['inactive'];
 
         session()->flash('contest_' . $competition->contest_id . '-' . 'competition_' . $competition->id . '-leaderboard', $list);
 
-        return view('competitions.leaderboard', compact('competition', 'leaderboard', 'pending'));
+        return view('competitions.leaderboard', compact('competition', 'leaderboard', 'pending', 'flagged'));
     }
 }
--- a/app/Services/Catalog.php
+++ b/app/Services/Catalog.php
@@ -38,6 +38,8 @@ class Catalog
         // would not require additional requests and API calls since all in flash cache.
 
         if ($method === 'rank') {
+            $users = $this->removeFlaggedUsers($users);
+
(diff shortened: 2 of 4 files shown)


angaither:
@mikefantini we might need you and impact team to weigh in on this...when/where do we want to display flagged reportbacks? should we hide them? or make a new table today?

mikefantini:
@angaither:

- When: we only want to show flagged reportbacks to admins, never to users in the leaderboard email
- Where: I think at the bottom of the table unranked in red makes sense (basically like you have in #255 except below the ranked users)

image

Let me know if that makes sense or if you have any other questions.

angaither:
instead of leaving them in the ranking, it made more sense (and was a lot easier) to make them their own table. 
image

mikefantini:
Works for me @angaither.

How difficult is it to maintain the other columns as well from the previous table (i.e. # promoted, approved, excluded, flagged, pending)?

angaither:
@mikefantini not difficult, it's just extra calls to get every item that I didn't think we needed. since once one thing is flagged, the whole thing is... so I didn't think it mattered

deadlybutter:
_notices my reportback is flagged_

me and Test T. are legit members, I do not approve of this censorship

angaither:
confirming flagged rbs are not in the email
image

deadlybutter:
noticed the two functions you added in catalog arent being used anywhere? other than that 👍  :shipit:

deadlybutter:
🆒  is a good enough justification for me

sbsmith86:
💯

PR #267, opened 2016-04-28
Initial User test cases - It takes a village

weerd:
#### What's this PR do?
This PR (with lots of help from @DFurnes) sets up some initial User test cases for Gladiator using the user repository. It also adds the new `mock()` method to the `TestCase` class to help allow the use of mocks when testing.

@angaither also added some User API tests as well to help start getting some API test action in too!

Well, do a `composer dump-autoload` and then run `phpunit`! Any failed tests? Good, there shouldn't be any!

@angaither I moved the User API tests into their own `UserApiTest` case class. Figured it would help make things easier with test code for User web app vs User API.

Refs #25

---
@DFurnes  @DoSomething/gladiator


angaither:
👍  HELL YES

DFurnes:
👍 👏

PR #271, opened 2016-04-28
Updating waiting room ux/ui and some other fixes.

### LINKED ISSUE #258: Seeding is causing weird data because it creates competitions with no users
image

 3 Competitions were created for first Contest during seeding with `0` users, and all user are still in the waiting room. This is causing a bunch of weird issues as I'm working on user competition count data exposure.

weerd:
#### What's this PR do?
This PR updates a few of the WaitingRoom views to use similar UX/UI as other resource views in Gladiator. It adds missing errors for the waiting room split page and also includes a few other fixes and code cleanup.

This also comments out the Competitions Seeder because during the seeding process, we're not properly handling how the system typically creates Competitions, where they are created based on a set of users in a WaitingRoom and then once the Competition is created, the WaitingRoom is cleared. Instead, in the seeder, we just create some random, empty Competitions in the first Contest and that's it. This was throwing off data when working on Competition user count info, etc.

Make sure the WaitingRoom related forms all work!

Fixes #258

---
@DoSomething/gladiator

### DIFF
--- a/app/Http/Controllers/WaitingRoomsController.php
+++ b/app/Http/Controllers/WaitingRoomsController.php
@@ -105,7 +105,7 @@ class WaitingRoomsController extends Controller
     {
         $room->fill($request->all())->save();
 
-        return redirect()->back()->with('status', 'Waiting room has been updated!');
+        return redirect()->route('waitingrooms.show', $room->id)->with('status', 'Waiting room has been updated!');
     }
 
     /**
--- a/resources/views/waitingrooms/partials/_form_waitingrooms.blade.php
+++ b/resources/views/waitingrooms/partials/_form_waitingrooms.blade.php
@@ -2,7 +2,7 @@
 
 <div class="form-item -padded">
     {{ Form::label('contest_id', 'Contest ID:', ['class' => 'field-label']) }}
-    {{ Form::text('contest_id', NULL, ['class' => 'text-field']) }}
+    {{ Form::text('contest_id', NULL, ['class' => 'text-field', 'readonly' => 'readonly']) }}
 </div>
 
 <div class="form-item -padded">
--- a/resources/views/waitingrooms/show.blade.php
+++ b/resources/views/waitingrooms/show.blade.php
@@ -9,26 +9,39 @@
 
     <div class="container">
         <div class="wrapper">
-            <div class="container__block">
-                <p>Contest ID: <a href="{{ route('contests.show', $room->contest_id) }}">{{ $room->contest_id }}</a></p>
-                <p>Campaign ID: {{ $contest->campaign_id }}</p>
-                <p>Campaign Run ID: {{ $contest->campaign_run_id }}</p>
-                <p>Signup Start Date: {{ $room->signup_start_date->format('F d, Y') }}</p>
-                <p>Signup End Date: {{ $room->signup_end_date->format('F d, Y') }}</p>
+            <div class="container__block -half">
+                <h2 class="heading">Information</h2>
 
+                <ul>
+                    <li><strong>Campaign ID:</strong> {{ $contest->campaign_id }}</li>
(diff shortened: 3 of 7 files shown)


resources/views/waitingrooms/partials/_form_waitingrooms.blade.php
```diff
@@ -2,7 +2,7 @@
 
 <div class="form-item -padded">
     {{ Form::label('contest_id', 'Contest ID:', ['class' => 'field-label']) }}
-    {{ Form::text('contest_id', NULL, ['class' => 'text-field']) }}
+    {{ Form::text('contest_id', NULL, ['class' => 'text-field', 'readonly' => 'readonly']) }}
```
weerd: This keeps the input from being edited. We never want the admin to actually edit this field or shit be broke.

app/Http/Controllers/WaitingRoomsController.php
```diff
@@ -105,7 +105,7 @@ public function update(WaitingRoomRequest $request, WaitingRoom $room)
     {
         $room->fill($request->all())->save();
 
-        return redirect()->back()->with('status', 'Waiting room has been updated!');
+        return redirect()->route('waitingrooms.show', $room->id)->with('status', 'Waiting room has been updated!');
```
deadlybutter: THANK YOU

resources/views/waitingrooms/partials/_form_waitingrooms.blade.php
deadlybutter: dumb question, why does it need to be readonly => readonly (verses just 'readonly')

resources/views/waitingrooms/show.blade.php
```diff
@@ -9,26 +9,39 @@
...
-                        <a href="{{ route('waitingrooms.export', $room->id) }}" class="button">Export</a>
-                    </li>
+        </div>
+    </div>
+
+    <div class="container">
+        <div class="wrapper">
+            <div class="container__block">
+            <h2 class="heading">Data export</h1>
+                <ul class="list">
```
deadlybutter: should this still be in a list pattern like this? We're only going to have this 1 option / list item

resources/views/waitingrooms/partials/_form_waitingrooms.blade.php
weerd: I wasn't 100% sure how blade would handle the attribute if no value is added (needs to be passed in to the input as an associative array). Technically the HTML5 properties like `disabled`, `readonly`, etc, which can exist as just stand alone attributes without the `=`, also work as `disabled = "disabled"` and `readonly = "readonly"` so it doesn't hurt it; just a bit redundant.

weerd: But yeah, not a dumb question 😉  Totally not obvious off the bat. Not sure if I pass an empty string to the value if it would make it `readonly = ""` or just output `readonly`, but wasn't too concerned to try.

resources/views/waitingrooms/show.blade.php
weerd: Yeah, 1 list item is fine. I'm following how it's set up other places. Could be debated, but I think eventually we're hoping to remove all the exports or minimize their use once emails are working 100%.

resources/views/waitingrooms/partials/_form_waitingrooms.blade.php
deadlybutter: not a big deal, just curious cause as you said some of the other html properties dont require it

resources/views/waitingrooms/show.blade.php
deadlybutter: kool

deadlybutter:
🚢 🇮🇹 

giphy 3

PR #277, opened 2016-04-28
Update test emails and refactor

### LINKED ISSUE #274: Update send to for test emails
- always send test emails to the user who is currently logged in and pressing the button
- don't grab the whole list of users for a test email

sbsmith86:
#### What's this PR do?

Refactors how I was handling test. Now, if it is a test email, we set the `users` resource to be the authenticated user..so only one messaged is processed and that messaged is processed with the admins information. Before, we were using the contest `sender_email` which 1) that might not be a real northstar account and 2) meant we had to still process messages for all users and pick on to send to the `sender_email`. This is better and fast.

Also cleaned up some code, added comments were I had missed it. 

Fixes #274

### DIFF
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -10,7 +10,7 @@ use Gladiator\Events\QueueMessageRequest;
 use Gladiator\Repositories\MessageRepository;
 use Gladiator\Repositories\UserRepositoryContract;
 use Gladiator\Services\Manager;
-use Gladiator\Services\Registrar;
+use Auth;
 
 class MessagesController extends Controller
 {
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/MessagesController.php
```diff
@@ -95,11 +96,19 @@ public function sendMessage(Message $message)
...
+            $user = Auth::user();
+            $user = $this->userRepository->find($user->id);
+            $user = $this->manager->appendReportback($user, null);
 
-        // Only send checkin messages to users who haven't reported back.
-        if ($message->type === 'checkin') {
-            $users = $users->where('reportback', null);
+            $users = [$user];
+        } else {
+            $users = $this->manager->getModelUsers($competition, true);
```
sbsmith86: Actually, don't review this yet! I realized i'm doing this twice on non-test emails now. Once, to get the users to send to and another time to generate the leaderboard. Gonna think this through a bit more but I think @weerd is pushing up some work that will allow me to pull competition activity from the competition so I don't have to generate the leaderboard manually for the email.

PR #278, opened 2016-04-28
Competition data exposure

### LINKED ISSUE #189: Add more info to competitions
It would be great to have the following information calculated per competition:

- [x] # of ppl who reported back
- [x] reportback rate: # of ppl who reported back/# of ppl in competition (or contest)
- [x] non-flagged impact quantity

weerd:
#### What's this PR do?
This PR updates the Competition show page to include a bunch of statistics for the competition. It also does some behind the scenes magic to keep the data snappy by flashing the session and re-flashing the session as needed so that viewing the leaderboard and exporting it aren't slow when including all the stats in the competition page.

image

Load up the Competition show page and see if all works ok, the new "Statistics" section appears and the viewing the Leaderboard and exporting it are still fast.

Fixes #189

---
@DoSomething/gladiator @DFurnes

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -47,13 +47,19 @@ class CompetitionsController extends Controller
      */
     public function show(Competition $competition)
     {
-        $contest = Contest::find($competition->contest_id);
+        $key = generate_model_flash_session_key($competition, ['includeActivity' => true]);
 
-        $contest = $this->manager->appendCampaign($contest);
+        if (session()->has($key)) {
+            $competition = session($key);
 
-        $users = $this->manager->getModelUsers($competition);
+            session()->reflash();
+        } else {
+            $competition = $this->manager->getCompetitionOverview($competition, true);
+        }
 
-        return view('competitions.show', compact('competition', 'contest', 'users'));
+        $statistics = $this->manager->getStatisticsForCompetition($competition);
+
+        return view('competitions.show', compact('competition', 'statistics'));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


angaither:
add some commas to that number! and reportback noun and verb it!

weerd:
ooo, good point 🤔

weerd:
updated!

image

app/helpers.php
```diff
@@ -84,6 +84,28 @@ function format_timestamp_for_display($timestamp)
...
+ * Generate a unique key for flashing model data to the session.
+ *
+ * @param  \Illuminate\Database\Eloquent\Model  $model
+ * @param  array  $options
+ * @return string
+ */
+function generate_model_flash_session_key($model, $options = [])
+{
+    if ($model instanceof \Gladiator\Models\Competition) {
+        $key = 'contest_' . $model->contest_id . '-' . 'competition_' . $model->id;
```
deadlybutter: dumb question, but do we need the contest id in this? shouldn't competition id's be unique of?

deadlybutter:
🆒  :shipit:

app/Http/Controllers/CompetitionsController.php
```diff
@@ -165,17 +176,21 @@ public function message(Competition $competition, Contest $contest)
...
     {
-        $competition = $competition->load('contest');
+        $key = generate_model_flash_session_key($competition, ['includeActivity' => true]);
 
-        $users = $this->manager->getModelUsers($competition, true);
+        if (session()->has($key)) {
+            $competition = session($key);
 
-        $list = $this->manager->catalogUsers($users);
+            session()->reflash();
```
angaither: re-⚡

angaither:
🚢

app/helpers.php
weerd: Yeah, that's true. Felt it wouldn't hurt to just be a bit more specific with the keys to avoid conflicts and overriding stuff. :shrug:

app/Http/Controllers/CompetitionsController.php
weerd: Yas! Found out about it in the docs. Pretty cool that it can be reflashed 💃 ⚡

PR #285, opened 2016-05-02
Retroactivly add peeps

### LINKED ISSUE #254: Retroactively add people in here
unclear if people should go into the waiting room or into a competition.

angaither:
#### What's this PR do?
- created a `+ user` in the contest view
- allows for a user to be added to an existing waiting room, or a competition
- added a new signup form to add the people in 

pull down, go to any competition, pick `+ user` and see them be added into the fun. 

I feel like this chunk could be cleaner
```
// Add the user to a competition or waiting room, but only if not already in it.
if ($request->has('competition_id') && ! $user->competitions()->where('id', $request->competition_id)->first()) {
    $user->competitions()->attach($request->competition_id);
} elseif ($request->has('waitingroom_id') && ! $user->waitingrooms()->where('id', $request->waitingroom_id)->first()) {
    $user->waitingRooms()->attach($request->waitingroom_id);
}
```

Fixes #254

### DIFF
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -169,9 +169,9 @@ class ContestsController extends Controller
 
         // Add the user to a competition or waiting room, but only if not already in it.
         if ($request->has('competition_id')) {
-            $this->manager->addUserToRoom('competitions', $request->competition_id, $user);
+            $this->manager->addUserToModel('competitions', $request->competition_id, $user);
         } else {
-            $this->manager->addUserToRoom('waitingrooms', $request->waitingroom_id, $user);
+            $this->manager->addUserToModel('waitingrooms', $request->waitingroom_id, $user);
         }
 
         return redirect()->route('contests.show', $contest->id)->with('status', 'Allllllright, we added that late punk!');
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -167,10 +167,10 @@ class ContestsController extends Controller
         }
 
         // Add the user to a competition or waiting room, but only if not already in it.
-        if ($request->has('competition_id') && ! $user->competitions()->where('id', $request->competition_id)->first()) {
-            $user->competitions()->attach($request->competition_id);
-        } elseif ($request->has('waitingroom_id') && ! $user->waitingrooms()->where('id', $request->waitingroom_id)->first()) {
-            $user->waitingRooms()->attach($request->waitingroom_id);
+        if ($request->has('competition_id')) {
+            $this->manager->addUserToRoom('competitions', $request->competition_id, $user);
+        } else {
+            $this->manager->addUserToRoom('waitingrooms', $request->waitingroom_id, $user);
         }
 
         return redirect()->route('contests.show', 1)->with('status', 'Allllllright, we added that late punk!');
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -155,10 +155,11 @@ class ContestsController extends Controller
     /**
      * The post request to add a user retroactively.
      *
+     * @param  \Gladiator\Models\Contest $contest
      * @param  SignupUserRequest $request
      * @return \Illuminate\Http\Response
      */
-    public function signupUser(SignupUserRequest $request)
+    public function signupUser(Contest $contest, SignupUserRequest $request)
     {
         $user = $this->registrar->findUserAccount($request->all());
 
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/ContestsController.php
```diff
@@ -134,4 +138,33 @@ public function export(Contest $contest)
...
+     * @return  \Illuminate\Http\Response
+     */
+    public function signupForm(Contest $contest)
+    {
+        $contest = $contest->load(['waitingroom', 'competitions']);
+
+        return view('contests.signup', compact('contest'));
+    }
+
+    public function signupUser(SignupUserRequest $request)
```
weerd: docblocks!

resources/views/contests/show.blade.php
```diff
@@ -27,6 +27,9 @@
                     <li>
                         <a href="{{ route('split', $contest->waitingRoom->id) }}" class="button -secondary">Split</a>
                     </li>
+                    <li>
+                        <a href="{{ route('contests.user.add', $contest->id) }}" class="button -secondary">+ User</a>
```
weerd: Maybe call the button "Signup User" since we already have a "Add User" in the user crud to avoid confusion?

angaither: i had that first and it ruins everything
image

weerd: We can fix visuals. For example, moving the export to it's own section like I've down in a few places now.. just didn't get to all of them.

angaither: 🎨 okie

app/Http/Controllers/ContestsController.php
```diff
@@ -134,4 +138,41 @@ public function export(Contest $contest)
...
+    {
+        $user = $this->registrar->findUserAccount($request->all());
+
+        if (! $user instanceof User) {
+            $user = $this->registrar->createUser($user);
+        }
+
+        // Add the user to a competition or waiting room, but only if not already in it.
+        if ($request->has('competition_id') && ! $user->competitions()->where('id', $request->competition_id)->first()) {
```
weerd: Yeah, this is tricky. 

```php
if ($request->has('competition_id')) {
  $this->manager->addUserToCompetition($user, $request->competition_id);
}
```
If the user already exists in that competition that method handles that logic?

And then something similar for the waiting room?

angaither: I'll move it to a manager, but it doesn't fail. if it finds the user it just silently ignores  the request and shows the success message, since the person was already in the waiting room/comp

weerd: If it finds the user using:

```php
$user->competitions()->where('id', $request->competition_id)->first()
```
? Wouldn't that just return the result if it finds it, otherwise `null`?

angaither: yup

weerd: So you could do something like:

```php

public function addUserToCompetition($user, $competitionId)
{
  $user = $user->competitions()->where('id', $competition_id)->first();

  if ($user) {
    throw new UserExistsInCompetitionException;
  } else {
    // add user and return user to calling function.
  }
}
```

then you can catch the exception in the calling function.  If this feels too over the top, maybe just leave what you had before. This could be useful if we foresee using these methods in other scenarios in the future.

angaither:
@weerd kewl, I updated it and was able to use the same method for both!

app/Services/Manager.php
```diff
@@ -418,4 +418,22 @@ public function appendReportbackItemToMessage($reportback_id, $reportback_item_i
...
+
+    /**
+     * Attach a user object to a waiting room or a competition.
+     *
+     * @param string $type
+     * @param string $id
+     * @param object $user
+     * @return bool
+     */
+    public function addUserToRoom($type, $id, $user)
```
weerd: Nice! Maybe call it `addUserToModel` and switch `$type` to `$model`. "room" is going to easily be confused with just the waiting room.

weerd:
¡OLÉ!
:sheep: it!

app/Http/Controllers/ContestsController.php
```diff
@@ -173,6 +174,6 @@ public function signupUser(SignupUserRequest $request)
             $this->manager->addUserToRoom('waitingrooms', $request->waitingroom_id, $user);
         }
 
-        return redirect()->route('contests.show', 1)->with('status', 'Allllllright, we added that late punk!');
```
weerd: Haha! Forgot you hardcoded it, lol.

PR #286, opened 2016-05-02
Moves the welcome email testing to the contest page

### LINKED ISSUE #241: Testing welcome email

deadlybutter:
#### What's this PR do?
Moves the welcome email test/send from the normal messages table to the contest page

Can you still send test emails for the welcome message?
Do other emails still send correctly?

gonna leave PR comments in a sec

Fixes #241

### DIFF
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -134,7 +134,7 @@ class MessagesController extends Controller
         // Kick off email sending
         event(new QueueMessageRequest($resources));
 
-        if (!$competitionId) {
+        if (! $competitionId) {
             return redirect()->back()->with('status', 'Fired that right the hell off!');
         }
 
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -92,7 +92,7 @@ class MessagesController extends Controller
     {
         // Get competition with activity.
         $competitionId = request('competition_id');
-        $competition = null;
+        // $competition = null;
 
         if ($competitionId) {
             $competition = Competition::find($competitionId);
@@ -126,7 +126,7 @@ class MessagesController extends Controller
 
         $resources = [
             'message' => $message,
-            'competition' => $competition,
+            'competition' => isset($competition) ? $competition : null,
             'users' => $users,
             'test' => request('test'),
         ];
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -92,16 +92,20 @@ class MessagesController extends Controller
     {
         // Get competition with activity.
         $competitionId = request('competition_id');
-        $competition = Competition::find($competitionId);
+        $competition = null;
 
-        // Get competition with activity from flash if it is there.
-        // Otherwise, grab it.
-        $key = generate_model_flash_session_key($competition, ['includeActivity' => true]);
+        if ($competitionId) {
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/MessagesController.php
```diff
@@ -92,16 +92,20 @@ public function sendMessage(Message $message)
     {
         // Get competition with activity.
         $competitionId = request('competition_id');
-        $competition = Competition::find($competitionId);
+        $competition = null;
```
deadlybutter: had to set this to NULL otherwise the code below errors out because $competition doesnt exist

https://github.com/DoSomething/gladiator/blob/master/app/Http/Controllers/MessagesController.php#L119

app/Http/routes.php
```diff
@@ -30,7 +30,7 @@
     Route::get('competitions/{competitions}/messages/{contest}', 'CompetitionsController@message')->name('competitions.message');
     Route::get('competitions/{competitions}/leaderboard', 'CompetitionsController@leaderboard')->name('competitions.leaderboard');
 
-    // Competitions
+    // Contests
```
deadlybutter: not related to the PR but was confusing me

app/Http/Controllers/MessagesController.php
sbsmith86: interesting why does that code even run when you are sending welcome emails? it should only run when the `type === 'checkin'`

app/Http/routes.php
sbsmith86: Good catch!

app/Http/Controllers/MessagesController.php
deadlybutter: no idea, i guess the compiler freaked out?

deadlybutter: is that even a thing in PHP?

sbsmith86: Also this is a test, so it shouldn't even get there. Since you are already checking if the competition ID was sent in the request, I think you can just check if the competition exists in the when setting that property in the `resources` array.

```
        $resources = [
            'message' => $message,
            'competition' => isset($competition) ? $competition : null,
            'users' => $users,
            'test' => request('test'),
        ];
```

deadlybutter: well it does get evaluated somehow. again I dont think that code is being executed, PHP is just confused because $competition was never created

screen shot 2016-05-02 at 2 46 07 pm

deadlybutter: oh wait im a doofus, @sbsmith86 youre right

app/Http/Controllers/MessagesController.php
```diff
@@ -92,16 +92,20 @@ public function sendMessage(Message $message)
     {
         // Get competition with activity.
         $competitionId = request('competition_id');
-        $competition = Competition::find($competitionId);
+        // $competition = null;
```
sbsmith86: think you can just delete this now :)

deadlybutter: im just full of :facepalm: today

sbsmith86:
Looks good :shipit:

Issue #280, opened 2016-04-29
featured reportback should be per competition


angaither:
- currently this is per contest

People have to change the featured RB every single time they send out a message, which is annoying. 
But it's also annoying to try to think of an _easy_ way to set the featured RB per competition. 
I guess we can't all win, like Queen 🐝  does.

![](http://rack.2.mshcdn.com/media/ZgkyMDEzLzAyLzA0L2Y4L0JleW9uY2VTdXBlLjZhMTgwLmdpZgpwCXRodW1iCTEyMDB4OTYwMD4/c69efe61/950/Beyonce-Super-Bowl-Half-Time-GIF.gif)

angaither:
I pushed up a [branch with the migrations](https://github.com/DoSomething/gladiator/tree/move-featured-reportback) and handed off to @deadlybutter

PR #289, opened 2016-05-03
Messages settings

### LINKED ISSUE #250: Expose default email config
Instead of dealing with updating default messages in a [file](https://github.com/DoSomething/gladiator/blob/master/app/Http/Utilities/Correspondence.php) this should be an admin config form in a settings menu.

It should update the defaults of any new messages created, but I think for ease of build, it shouldn't retroactivly update already created messages.

weerd:
#### What's this PR do?
This PR adds a new section for Settings and within that section there is a sub section for Default Messages Settings. It allows editing the default messages for when additional Contests are created.

`composer dump-autoload` reset the DB, re-migrate, and re-seed. Check that there's a new **messages_settings** table. Then login and go to Settings -> Messages and try editing the default messages. Next create a new Contest and see if the the edits to the defaults that are populated into the form are what you had previously changed.

I attempted to create a generalized Settings system, with the intent that it could become a separate DS PHP package that we could pull into multiple projects. However, the complexity of Messages, with their multipart fields (subject, body, pro_tip, etc) made that approach too convoluted. Scrapped it but kept the code for a potential future implementation. 

While some of the code in this PR isn't ideal and could very well be refactored a bit, I wanted to get something up that works given I lost a bunch of time with the original approach.

Fixes #250

---
@DoSomething/gladiator


DFurnes:
@weerd Go back to bed!! 😷

app/Models/MessageSetting.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+class MessageSetting extends Model
+{
+    /**
+     * Get the fillable attributes for the model.
+     *
+     * @return array
+     */
+    protected $fillable = ['type', 'key', 'label', 'subject', 'body', 'pro_tip', 'signoff', 'shoutout'];
```
angaither: @deadlybutter just a note here, you'll want to pull shoutout out of here when you make the pr for #280

database/migrations/2016_05_03_040602_create_messages_settings_table.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        Schema::create('messages_settings', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('type')->index();
+            $table->integer('key')->unsigned();
+            $table->string('label')->nullable();
+            $table->string('subject')->nullable();
+            $table->longText('body')->nullable();
+            $table->longtext('pro_tip')->nullable();
+            $table->longText('shoutout')->nullable();
```
angaither: here as well @deadlybutter

app/Models/MessageSetting.php
deadlybutter: 👍

angaither:
I tested and it works, going to merge this 🎉

Issue #292, opened 2016-05-03
welcome email issues 


angaither:
seeing a lot of these

```
2016-05-03 20:47:14] production.ERROR: exception 'Swift_RfcComplianceException' with message 'Address in mailbox given [] does not comply with RFC 2822, 3.6.2.' in /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php:348
Stack trace:
#0 /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php(263): Swift_Mime_Headers_MailboxHeader->_assertValidAddress('')
#1 /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php(106): Swift_Mime_Headers_MailboxHeader->normalizeMailboxes(Array)
#2 /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php(63): Swift_Mime_Headers_MailboxHeader->setNameAddresses(Array)
#3 /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/SimpleHeaderFactory.php(58): Swift_Mime_Headers_MailboxHeader->setFieldBodyModel(Array)
#4 /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/SimpleHeaderSet.php(68): Swift_Mime_SimpleHeaderFactory->createMailboxHeader('To', Array)
#5 /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/SimpleMessage.php(323): Swift_Mime_SimpleHeaderSet->addMailboxHeader('To', Array)
#6 /var/www/gladiator/releases/20160502193114/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/SimpleMessage.php(299): Swift_Mime_SimpleMessage->setTo(Array)
#7 /var/www/gladiator/releases/20160502193114/vendor/laravel/framework/src/Illuminate/Mail/Message.php(137): Swift_Mime_SimpleMessage->addTo(NULL, NULL)
[...]

PR #294, opened 2016-05-03
Email queue bugs

sbsmith86:
#### What's this PR do?

Fixes some errors we were seeing in the logs when emails were failing. 

* The first issue was that, while we are validating emails and checking that an email exists before processing a message for a user, I was not taking into account a situation like the welcome email, where there is only one user. So, if the email wasn't valid in that case, no message was processed or added to the `allMessages` array, and we were still trying to loop through `allMessages` which was causing an error around that `foreach`. [Now we make sure there are messages before trying to send](https://github.com/DoSomething/gladiator/compare/email-queue-bugs?expand=1#diff-90614d9e91b9ac6b77756a5777299189R40)

* Second issue was around logging. We needed to 1) [log emails that were not valid from our validation](https://github.com/DoSomething/gladiator/compare/email-queue-bugs?expand=1#diff-4f39f258033e5d77d4d966ae87da96c2R166), and 2) log emails when they [fail to send](https://github.com/DoSomething/gladiator/compare/email-queue-bugs?expand=1#diff-90614d9e91b9ac6b77756a5777299189R69)

This will mostly be helpful in prod, however, if you have a competition with users that bad, if you pull down this branch and send to all users you will see messages in the log with useful messages like.

```
[2016-05-03 22:12:33] local.ALERT: User email is not valid {"id":"557775e9a59dbf3b7a8b457b","email":null} 
``` 

These two additions will one, make sure that the queue and logs don't get clogged up with a bunch of failed attempts, but also allow us to debug the issue better. 

For example, it is odd to me that people are getting welcome emails to email addresses that are not valid cause they should be logged in users in Phoenix at that time, and therefore valid ¯\_(ツ)_/¯

Addresses #292

### DIFF
--- a/app/Listeners/QueueMessage.php
+++ b/app/Listeners/QueueMessage.php
@@ -65,7 +65,7 @@ class QueueMessage implements ShouldQueue
                 $msg->from($settings['from'], $settings['from_name']);
 
                 $msg->to($settings['to'], $settings['to_name'])->subject($settings['subject']);
-            } catch(Swift_RfcComplianceException $e) {
+            } catch (Swift_RfcComplianceException $e) {
                 Log::alert('Message failed to send', ['error' => $e]);
             }
         });
--- a/app/Listeners/QueueMessage.php
+++ b/app/Listeners/QueueMessage.php
@@ -3,6 +3,7 @@
 namespace Gladiator\Listeners;
 
 use Log;
+use Swift_RfcComplianceException;
 use Illuminate\Mail\Mailer;
 use Gladiator\Events\QueueMessageRequest;
 use Illuminate\Contracts\Queue\ShouldQueue;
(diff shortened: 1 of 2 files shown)


app/Listeners/QueueMessage.php
```diff
@@ -35,16 +37,18 @@ public function handle(QueueMessageRequest $event)
...
+                $settings = [
+                    'subject' => $content['message']['subject'],
+                    'from' => $email->contest->sender_email,
+                    'from_name' => $email->contest->sender_name,
+                    'to' => $content['user']->email,
+                    'to_name' => $content['user']->first_name,
+                ];
 
-            $this->sendMail($content['message'], $settings);
+                $this->sendMail($content['message'], $settings);
```
angaither: I'm still a bit confused by this, do welcome emails still send? 
if we are sending a one off email to just one user, don't we need to check [here](https://github.com/DoSomething/gladiator/blob/master/app/Services/Email.php#L154)?
won't this  `sendMail` call not fire because it's within the for loop if there are multiple messages?

sbsmith86: The `Email` class takes in an array of users to send messages to. In the case of a welcome email it passed an array of only one user. The code you point to, goes through each user passed, checks if they have valid email address and then adds it to an `allMessages` property that we loop through here. 

The issue that I am trying to fix here is when that one user passed, doesn't have a valid email, so no `allMessage` property is passed and we don't want to try to loop through it. 

So welcome emails will still send, when the user is valid, cause we will build a message for them. I added a line [to log](https://github.com/DoSomething/gladiator/pull/294/files#diff-4f39f258033e5d77d4d966ae87da96c2R167) when a user doesn't pass validation so we can see the user that is failing.

angaither:
👍

PR #298, opened 2016-05-04
Welcome email bugs

sbsmith86:
#### What's this PR do?

Updated the welcome email logic so that it always grabs the full Northstar user before sending. We were seeing issues with depending on `$this->registrar->findUserAccount();` for the user account because if the account already exists in gladiator, it will just return the User object from the DB that doesn't have an email address attached to it. 

Now we pull in the User Repository and use that to grab the user before sending so we make sure we get all the user info. 

⚠️  I don't have a working dev version of phoenix on my local so I wasn't able to test this really haha. But hoping to get this on QA for further debugging.

Addresses #292

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -2,6 +2,7 @@
 
 namespace Gladiator\Http\Controllers\Api;
 
+use Log;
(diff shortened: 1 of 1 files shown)


angaither:
👍

PR #304, opened 2016-05-05
Welcome email take 2

sbsmith86:
#### What's this PR do?

* Created a failed jobs table migration that logs failed queue jobs after they have reached the limit of attempts **(now limited to 3 attempts instead of infinity, like before)**

* Creates a `sendEmail()` function to send one off emails and we use that in the api `UsersController` to send of the welcome emails. I think there were issues with us setting the `$user` variable as a northstar object before returning, so this separates the concerns better. 

Pull down branch, wipe and reset database, make a request to add a user and it should work, and that user should get a welcome email. 

Addresses #292

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -514,7 +514,7 @@ class Manager
             'test' => false,
         ];
 
-        Log::debug('Gladiator\Services\Manager -- Sending welcome email', ['user' => $user]);
+        Log::debug('Gladiator\Services\Manager -- Sending ' . $params['type'] . ' email', ['users' => $resources['users']]);
 
         event(new QueueMessageRequest($resources));
     }
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -502,9 +502,16 @@ class Manager
      * @param  Gladiator\Models\User     $user
      * @param  Gladiator\Models\Contest  $contest
      * @param  array $params
+     * @return null|void
      */
     public function sendEmail($user, $contest, $params)
     {
+        if (! isset($params)) {
+            Log::error('Gladiator\Services\Manager -- No params passed');
+
+            return null;
+        }
+
         $message = Message::where($params)->first();
 
         if ($message) {
@@ -521,6 +528,8 @@ class Manager
             event(new QueueMessageRequest($resources));
         } else {
             Log::error('Gladiator\Services\Manager -- Message not found', ['params' => $params]);
+
+            return null;
         }
     }
 }
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -502,9 +502,9 @@ class Manager
      * @param  Gladiator\Models\User     $user
      * @param  Gladiator\Models\Contest  $contest
      */
-    public function sendEmail($user, $contest)
+    public function sendEmail($user, $contest, $params)
     {
-        $message = Message::where(['contest_id' => $contest->id, 'type' => 'welcome'])->first();
+        $message = Message::where($params)->first();
 
         $resources = [
             'message' => $message,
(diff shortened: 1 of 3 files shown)


weerd:
> now limited to 3 attempt instead of infinity, like before

😱

app/Services/Manager.php
```diff
@@ -495,4 +495,27 @@ public function addUserToModel($model, $id, $user)
...
         return true;
     }
+
+    /**
+     * Fires off an event to send an email a user.
+     *
+     * @param  Gladiator\Models\User     $user
+     * @param  Gladiator\Models\Contest  $contest
+     */
+    public function sendEmail($user, $contest)
```
weerd: is this a general email sender method now? Maybe we also pass in the `$type` and `$key` so it can send other messages, not just the welcome. if that's the case then we'll need the `$key` in there as well to differentiate between emails. This could also be a later update? Maybe note it?

weerd: Maybe you pass in a `$parameters` var that is an array like, 

```php
[
  'type' => 'welcome',
  'key' => '1',
]
```
and you use those params in the `where` statement when grabbing the `$message`

weerd: also, just in case the message doesn't exist for whatever reason... maybe wrap the `$resources` and `event()` in an `if`

```php
if ($message) {
  ... // do all the sending stuff here since we found the message
}
```

app/Services/Manager.php
```diff
@@ -495,4 +495,27 @@ public function addUserToModel($model, $id, $user)
...
+
+    /**
+     * Fires off an event to send an email a user.
+     *
+     * @param  Gladiator\Models\User     $user
+     * @param  Gladiator\Models\Contest  $contest
+     */
+    public function sendEmail($user, $contest, $params)
+    {
+        $message = Message::where($params)->first();
```
weerd: Oh nice. Forgot you can just pass the array of params in 💃

weerd: should `$params = []` in the function def in case no params are passed? or maybe there's a check inside the method and returns out if no params passed? :shrug:

sbsmith86: hmm i feel like that would then cause the query to just return all messages. I feel like requiring it is important and allows the developer to easily see where something failed. I did add the check to see if a message was found our not with a log message that says what params were passed, which could be useful.

app/Services/Manager.php
```diff
@@ -495,4 +495,32 @@ public function addUserToModel($model, $id, $user)
...
         return true;
     }
+
+    /**
+     * Fires off an event to send an email a user.
+     *
+     * @param  Gladiator\Models\User     $user
+     * @param  Gladiator\Models\Contest  $contest
+     * @param  array $params
+     */
```
weerd: add a `@return void` :)

app/Services/Manager.php
```diff
@@ -495,4 +495,41 @@ public function addUserToModel($model, $id, $user)
...
+     *
+     * @param  Gladiator\Models\User     $user
+     * @param  Gladiator\Models\Contest  $contest
+     * @param  array $params
+     * @return null|void
+     */
+    public function sendEmail($user, $contest, $params)
+    {
+        if (! isset($params)) {
+            Log::error('Gladiator\Services\Manager -- No params passed');
```
weerd: We could eventually throw an exception here instead of just logging. One that is caught by the Handler.php but doesn't break out of the processing! But not necessary for merge. Just noting :)

weerd:
¡OLÉ!
:sheep: it!

app/Services/Manager.php
sbsmith86: @weerd True! the logging isn't really useful since we truncate the log file on these servers, but is helpful for QA in the meantime. I would really like us to get that development hub up at some point. :)

PR #305, opened 2016-05-05
Use message class in manager class

sbsmith86:
#### What's this PR do?
Import the message class in to the Manager class so we can actually use it. 

```
[2016-05-05 19:01:17] production.ERROR: exception 'Symfony\Component\Debug\Exception\FatalErrorException' with message 'Class 'Gladiator\Services\Message' not found' in /var/www/gladiator/releases/20160505184844/app/Services/Manager.php:515
```

#292

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -2,6 +2,7 @@
 
 namespace Gladiator\Services;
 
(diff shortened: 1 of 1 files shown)


angaither:
ya ya 🚢

PR #306, opened 2016-05-05
Import classes

sbsmith86:
#### What's this PR do?

* Import classes i'm using in the Manager class

* Remove logging since it will log for every welcome email and that would be a lot of logging

#292

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -2,12 +2,14 @@
 
 namespace Gladiator\Services;
 
+use Log;
 use Gladiator\Models\Message;
 use Gladiator\Models\Contest;
 use Gladiator\Repositories\CacheCampaignRepository;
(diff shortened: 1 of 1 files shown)


app/Services/Manager.php
```diff
@@ -524,12 +523,8 @@ public function sendEmail($user, $contest, $params)
                 'test' => false,
             ];
 
-            Log::debug('Gladiator\Services\Manager -- Sending ' . $params['type'] . ' email', ['users' => $resources['users']]);
-
             event(new QueueMessageRequest($resources));
         } else {
-            Log::error('Gladiator\Services\Manager -- Message not found', ['params' => $params]);
```
sbsmith86: Or maybe I should leave the error logs? they will be less (hopefully)?

angaither:
👍

PR #311, opened 2016-05-06
Adds old helper to split form where applicable

### LINKED ISSUE #295: split form should save input
when you're filling out the split form, and you get an error, the other input you added is remove. 

I filled out the form, and the date when missing, when I only didn't add rules.
image

deadlybutter:
#### What's this PR do?
adds `old` support where possible

Cause an error on the page, does it save some input?

Not sure how to use Old for the date input or our custom selector for the day.

Fixes #295

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -55,12 +55,12 @@ function correspondence($message = null, $field = null)
  * @param  string  $defaut
  * @return string
  */
-function format_date_form_field($model, $field, $defaut = 'MM/DD/YYYY')
+function format_date_form_field($model, $field, $default = 'MM/DD/YYYY')
 {
     $oldDate = old($field);
 
     if ($oldDate) {
-        return $oldDate;
+        return (new Carbon($oldDate))->format('Y-m-d');
     }
 
     $attribute = $model->getAttribute($field);
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -60,7 +60,7 @@ function format_date_form_field($model, $field, $default = 'MM/DD/YYYY')
     $oldDate = old($field);
 
     if ($oldDate) {
-        return (new Carbon($oldDate))->format('Y-m-d');
+        return $oldDate;
     }
 
     $attribute = $model->getAttribute($field);
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -52,7 +52,7 @@ function correspondence($message = null, $field = null)
  *
  * @param  \Illuminate\Database\Eloquent\Model  $model
  * @param  string  $field
- * @param  string  $defaut
+ * @param  string  $default
  * @return string
  */
 function format_date_form_field($model, $field, $default = 'MM/DD/YYYY')
--- a/resources/views/waitingrooms/partials/_form_split.blade.php
+++ b/resources/views/waitingrooms/partials/_form_split.blade.php
@@ -6,17 +6,17 @@
     <div class="container__block">
         <div class="form-item -padded">
             <label class="field-label" for="competition_end_date">End date for these competitions:</label>
-            <input type="date" name="competition_end_date" id="competition_end_date" value='MM/DD/YYYY' class="text-field"></input>
+            <input type="date" name="competition_end_date" id="competition_end_date" value='MM/DD/YYYY' class="text-field" value="{{ old('competition_end_date', 'MM/DD/YYYY') }}"></input>
         </div>
 
         <div class="form-item -padded">
             <label class="field-label" for="id">Max users per competition:</label>
-            <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" }/>
+            <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" value="{{ old('competition_max') }}" />
         </div>
 
         <div class="form-item -padded">
             <label class="field-label" for="rules_url">Rules URL:</label>
-            <input type="text" name="rules_url" id="rules_url" class="text-field" placeholder="http://docs.google.com/the-rules" />
+            <input type="text" name="rules_url" id="rules_url" class="text-field" placeholder="http://docs.google.com/the-rules" value="{{ old('rules_url') }}" />
         </div>
 
         @include('competitions.partials._form_leaderboard_msg_day_field', [ 'default' => null ])
--- a/resources/views/waitingrooms/partials/_form_split.blade.php
+++ b/resources/views/waitingrooms/partials/_form_split.blade.php
@@ -11,7 +11,7 @@
 
         <div class="form-item -padded">
             <label class="field-label" for="id">Max users per competition:</label>
-            <input type="number" name="competition_max" id="competition_max" class="text-field" data-total="{{count($room->users)}}" placeholder=300 value="{{ old('competition_max') }}" />
+            <input type="number" name="competition_max" id="competition_max" class="text-field" data-total="{{count($room->users)}}" value="{{ old('competition_max', 300) }}" />
         </div>
 
         <div class="form-item -padded">
--- a/resources/views/waitingrooms/partials/_form_split.blade.php
+++ b/resources/views/waitingrooms/partials/_form_split.blade.php
@@ -11,7 +11,7 @@
 
         <div class="form-item -padded">
             <label class="field-label" for="id">Max users per competition:</label>
-            <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" value="{{ old('competition_max') }}" />
+            <input type="number" name="competition_max" id="competition_max" class="text-field" data-total="{{count($room->users)}}" placeholder=300 value="{{ old('competition_max') }}" />
         </div>
 
         <div class="form-item -padded">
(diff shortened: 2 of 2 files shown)


angaither:
> Not sure how to use Old for the date input or our custom selector for the day.

you should be able to still use `old` for those fields

deadlybutter:
screen shot 2016-05-09 at 11 06 37 am

the value is stored correctly (outputted in a paragraph above ^) but the input isnt taking the value correctly. @weerd do you know of a more proper way to do this?

weerd:
Yeah, there's somewhere, maybe the waiting room that I did a fix that puts the date in correctly when using old. I created a helper method to, well, _help_ with making it easier :)

weerd:
Lemme try and find where it is...

weerd:
Check the `_form_contest.blade.php`... and also look into where the data is coming from. I think I had to do something to it in the controller that passes the data to the view as well.

deadlybutter:
The function it calls `in the form you mentioned is format_date_form_field` but that simply returns the `old` value, which is what i was already doing.

below that however it does some formatting magic if its grabbing an attribute from the model. but the output of this format function is already identical to the output of `old`. so thats not the problem

under the input i have a paragraph outputting the format_date_form_field, and it looks correct 

```
            <input type="date" name="competition_end_date" id="competition_end_date" value='MM/DD/YYYY' class="text-field" value="{{ format_date_form_field($room, 'competition_end_date') }}"></input>
        </div>
        <p>{{ format_date_form_field($room, 'competition_end_date') }}</p>
```

screen shot 2016-05-09 at 11 45 29 am

however as you can see it keeps getting transformed into the generic yyyy mm dd string once it gets inserted in the input value. 

just documenting my findings so far, not sure why this is being so problematic.

sbsmith86:
@deadlybutter try taking it out of the quotes in the `value` field. I ran into this before and it had something to do with that.

sbsmith86:
From my convo with @weerd about something like this

image

weerd:
@deadlybutter can you actually use that helper function though. It does more than just grab the `old()` value if available. It also populates the field if value already exists on model if loading the edit page, etc.

sbsmith86:
more context...sorry for the flood of images 

image

deadlybutter:
@weerd right thats why i didnt suggest taking it out, but the purposes in this case the added functionality isnt needed (there is no competition to read an attribute from). regardless, i have no problem using the function.

i tried @sbsmith86 suggestion to remove the quotes but that didnt seem to do it either.

deadlybutter:
just tried manually changing the value in the input with the string format_date_form_field gave, worked fine. 

something strange is going on...

weerd:
yeah. I experienced that same weirdness and investigated a bunch. The helper function accounts for it 😉

deadlybutter:
i posted the code above which is using the helper function (and tried) to explain why its not fixing this https://github.com/DoSomething/gladiator/pull/311#issuecomment-217903538

weerd:
@deadlybutter It's probably because in the code you inserted in your [comment](https://github.com/DoSomething/gladiator/pull/311#issuecomment-217903538) you declare **two** `value` attributes instead of a `value` and a `placeholder` 😜

weerd:
HTML woes 😉

deadlybutter:
facepalm

deadlybutter:
@weerd updated

app/helpers.php
```diff
@@ -55,12 +55,12 @@ function correspondence($message = null, $field = null)
...
  * @return string
  */
-function format_date_form_field($model, $field, $defaut = 'MM/DD/YYYY')
+function format_date_form_field($model, $field, $default = 'MM/DD/YYYY')
 {
     $oldDate = old($field);
 
     if ($oldDate) {
-        return $oldDate;
+        return (new Carbon($oldDate))->format('Y-m-d');
```
weerd: so this isn't really needed. The "old" value would be whatever was in the input as the value prior, so there shouldn't be a need to re-carbonize and reformat it.

app/helpers.php
```diff
@@ -55,12 +55,12 @@ function correspondence($message = null, $field = null)
  * @param  string  $defaut
  * @return string
  */
-function format_date_form_field($model, $field, $defaut = 'MM/DD/YYYY')
+function format_date_form_field($model, $field, $default = 'MM/DD/YYYY')
```
weerd: good catch! Update the docblock one too, cause that's misspelled as well!

app/helpers.php
deadlybutter: how did this get pushed up, i deleted that

weerd: git always remembers, haha :shrug:

deadlybutter:
@weerd 💥

resources/views/waitingrooms/partials/_form_split.blade.php
```diff
@@ -6,17 +6,17 @@
...
         <div class="form-item -padded">
             <label class="field-label" for="competition_end_date">End date for these competitions:</label>
-            <input type="date" name="competition_end_date" id="competition_end_date" value='MM/DD/YYYY' class="text-field"></input>
+            <input type="date" name="competition_end_date" id="competition_end_date" class="text-field" value="{{ format_date_form_field($room, 'competition_end_date') }}"></input>
         </div>
 
         <div class="form-item -padded">
             <label class="field-label" for="id">Max users per competition:</label>
-            <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" }/>
+            <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" value="{{ old('competition_max') }}" />
```
weerd: Again, you have 2 `value` attributes...

weerd: use `placeholder` if you want an example... or use the `or` statement `{{ old('competition_max') or '300' }}`... I think something like that

deadlybutter: i must have copy pasted instead of cut pasted, also all of these work in testing so not sure why two values are working here

deadlybutter: hmm ok so i changed it to placeholder=300 and removed the duplicate value. tested again and things still work

resources/views/waitingrooms/partials/_form_split.blade.php
```diff
@@ -6,17 +6,17 @@
...
         <div class="form-item -padded">
             <label class="field-label" for="competition_end_date">End date for these competitions:</label>
-            <input type="date" name="competition_end_date" id="competition_end_date" value='MM/DD/YYYY' class="text-field"></input>
+            <input type="date" name="competition_end_date" id="competition_end_date" class="text-field" value="{{ format_date_form_field($room, 'competition_end_date') }}"></input>
         </div>
 
         <div class="form-item -padded">
             <label class="field-label" for="id">Max users per competition:</label>
-            <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" }/>
+            <input type="number" name="competition_max" id="competition_max" class="text-field" data-total="{{count($room->users)}}" placeholder=300 value="{{ old('competition_max') }}" />
```
weerd: You want it to actually have a value of `300` in the input though, right? Then you should probs just not use the `placeholder` and do the other thing I mentioned with the `or` keyword. I just mentioned `placeholder` as an example. Placeholders are used when you want to provide an example to user of what the input should look like... like an email example placeholder, etc.

deadlybutter: right but ive noticed lately chris & hillary havent been sticking to the 300 number, been making it more custom. which is fine but didnt want to force the 300 on them now that they've been changing it

weerd: Which is fine. They can still edit it. But it's otherwise weird that it could be left blank (cause your function that this submits to defaults to 300 anyhoo), but makes it less straightforward that it's a number that is necessary. Maybe we  keep the placeholder, but add a validation to that field to indicate it's required, apart from just being a number...?

deadlybutter: its already required, will show an error if you dont fill it out. but i updated the code to just have 300 as the default if old() doesnt return anything

resources/views/waitingrooms/partials/_form_split.blade.php
```diff
@@ -6,17 +6,17 @@
...
         <div class="form-item -padded">
             <label class="field-label" for="competition_end_date">End date for these competitions:</label>
-            <input type="date" name="competition_end_date" id="competition_end_date" value='MM/DD/YYYY' class="text-field"></input>
+            <input type="date" name="competition_end_date" id="competition_end_date" class="text-field" value="{{ format_date_form_field($room, 'competition_end_date') }}"></input>
         </div>
 
         <div class="form-item -padded">
             <label class="field-label" for="id">Max users per competition:</label>
-            <input type="number" name="competition_max" id="competition_max" class="text-field" value="300" data-total="{{count($room->users)}}" }/>
+            <input type="number" name="competition_max" id="competition_max" class="text-field" data-total="{{count($room->users)}}" value="{{ old('competition_max', 300) }}" />
```
weerd: nice!

weerd:
the saga ends... 👍

deadlybutter:
dont worry ill get michael bay to direct the next one 💥

weerd:
haha 😉

PR #293, opened 2016-05-03
Move featured reportback

deadlybutter:
#### What's this PR do?
Moves specifying the featured report-back outside the contest 

screen shot 2016-05-03 at 5 00 08 pm

#### How should this be manually tested?
I had to do some complex debugging (aka calling the relevant function getFeaturedReportback ) to test its pulling in correctly
Also verify the migrations work, seeds work, and the form properly updates the DB

#### Any background context you want to provide?
😪 

#### What are the relevant tickets?
Fixes #280

PR #314, opened 2016-05-07
Readding an accidentally deleted default in Correspondence.

weerd:
#### What's this PR do?
This PR re-adds a missing item in the default Correspondence array that was accidentally deleted in #293.

Does `artisan db:seed` still throw errors? No? Great!

🌵 

---
@DoSomething/gladiator


angaither:
👍

PR #315, opened 2016-05-09
Add checks to make sure we got a reportback from phoenix

### LINKED ISSUE #288: Fake featured reportback IDs cause emails to fail
If you set fake or wrong IDs as the Featured Reportbak in a leaderboard email, the email seems to fail. 

We should have a safety check that sends the email anyway.

sbsmith86:
#### What's this PR do?

Add checks to make sure we actually got a reportback back from phoenix. I was seeing issues with leaderboard emails when a featured reportback was set, but either the RID or Item id was bad. The email would fail cause it didn't check that it got a good response back. 

Add some invalid IDs to a leaderboard email and try to test. It should still send out but will not have the featured RB cause it ain't real or don't exist. 

Fixes #288 


angaither:
👍

PR #317, opened 2016-05-09
Updating waiting room edit form to not use Form facade.

weerd:
#### What's this PR do?
This PR does a little bit of form cleanup and removes the use of the `Form` facade in the WaitingRoom edit form.

Try editing a Waiting Room for a Contest and see if it breaks!

We're aiming to remove the use of the `Form` facade. While it used to be great for use in Laravel projects, there are times when it causes more problems, and it's typically better to just control the output of the HTML markup more directly.

🌵 

---
@DoSomething/gladiator


deadlybutter:
kool :shipit:

PR #319, opened 2016-05-17
Return view to list multiple search results

### LINKED ISSUE #318: Return multiple users from search
Allow search to return multiple users. Bad judgement call on my part, i figured there wouldn't be two people with the same ID, email, or phone. While that may be true, it seems search can sometimes strip out info so that we aren't always searching for the exact email, for example, so multiple people might be returned.

sbsmith86:
#### What's this PR do?

Returns a view the lists multiple users if more than one is found.

Search for a user, if there are multiple results you should see a view listing all the users. 

See [issue]( https://github.com/DoSomething/gladiator/issues/318 )
Fixes #318

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -188,6 +188,12 @@ class UsersController extends Controller
         // Attempt to fetch all users.
         $users = $this->manager->search($query);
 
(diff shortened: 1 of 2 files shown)


angaither:
👍

weerd:
👍

PR #322, opened 2016-06-13
zero index correspondence messages

### LINKED ISSUE #321: Zero-index message keys
There is [logic](https://github.com/DoSomething/gladiator/blob/master/resources/views/messages/partials/_reportback.blade.php) that depends on the different leaderboard messages being zero-indexed, but the settings messages and the default messages defined in the `Correspondence` class are one-indexed (? is that a thing? shrug ) so we need them to be zero-index so it is more inline with 💻  and the logic will work.

sbsmith86:
#### What's this PR do?

Zero-index message keys in the Correspondence class. 

See #321 

Fixes #321

### DIFF
--- a/app/Http/Utilities/Correspondence.php
+++ b/app/Http/Utilities/Correspondence.php
@@ -12,7 +12,7 @@ class Correspondence
     protected static $defaultMessages = [
         [
             'type' => 'welcome',
-            'key' => '1',
+            'key' => '0',
             'label' => 'Welcome',
             'subject' => 'Welcome, :first_name:!',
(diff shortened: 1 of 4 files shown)


weerd:
@sbsmith86 Do you think the indexing on the other items affect anything?

sbsmith86:
@weerd ya i checked... only this https://github.com/DoSomething/gladiator/blob/f4fda9c508118e9051a60c532bf5d7041d7487a1/app/Http/Controllers/Api/UsersController.php#L98-L102

weerd:
Coool :)

:sheep: it!

PR #324, opened 2016-06-28
Welcome message bug

### LINKED ISSUE #323: Welcome email bug
Welcome emails are not sending with customized content.

sbsmith86:
#### What's this PR do?
I found two problems.

1) Welcome test messages were f**ked cause we weren't passing in the contest explicity and it was causing errors when trying to grab information about the contest in the `Email` class.

2) Welcome emails were f**ked cause we pass in a `$params` array that we use to query the `Message` table but we were just looking for the first email of type welcome with key 0...

We probably didn't notice cause mostly all the welcome emails were the same..but if you updated one for a specific contest, we didn't care. 

Go to a contest, customize the welcome message.

Post to `/users` and you should get a welcome email with the right content. 

also go ahead and send yourself a test email. that should work as well. 

giphy 1

Fixes #323

### DIFF
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -92,7 +92,9 @@ class MessagesController extends Controller
     {
         // Get competition with activity.
         $competitionId = request('competition_id');
-        // $competition = null;
+
+        $contest = Contest::find(request('contest_id'));
+        $this->manager->appendCampaign($contest);
 
         if ($competitionId) {
             $competition = Competition::find($competitionId);
@@ -127,6 +129,7 @@ class MessagesController extends Controller
         $resources = [
             'message' => $message,
             'competition' => isset($competition) ? $competition : null,
+            'contest' => $contest,
             'users' => $users,
             'test' => request('test'),
         ];
--- a/app/Http/Controllers/MessagesController.php
+++ b/app/Http/Controllers/MessagesController.php
@@ -94,7 +94,7 @@ class MessagesController extends Controller
         $competitionId = request('competition_id');
 
         $contest = Contest::find(request('contest_id'));
-        $this->manager->appendCampaign($contest);
+        $contest = $this->manager->appendCampaign($contest);
 
         if ($competitionId) {
             $competition = Competition::find($competitionId);
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -99,6 +99,7 @@ class UsersController extends ApiController
         $params = [
             'type' => 'welcome',
             'key' => 0,
+            'contest_id' => $contest->id,
         ];
 
         $this->manager->sendEmail($user, $contest, $params);
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -93,7 +93,7 @@ class UsersController extends ApiController
         }
 
         $contest->waitingRoom->users()->attach($user->id);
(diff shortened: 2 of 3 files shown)


angaither:
@sbsmith86 nothing additional needs to be POSTed to the `/users` endpoint, correct? 
> Post to /users and you should get a welcome email with the right content.

sbsmith86:
@angaither ah nah you have to post some data: 

```
{ 
   "id"   : "INSERT_YOUR_EMAIL",
   "term" : "email",
   "campaign_id"   : CAMPAIGN_ID,
   "campaign_run_id" : RUN_ID
}
```

sbsmith86:
@angaither here is postman link to the collection i set up https://www.getpostman.com/collections/473d2791284c5e5b599f

angaither:
@sbsmith86 right right, but the data we have to post is the same as it was before. (just want to make sure nothing has to change on the Phoenix end)

sbsmith86:
@angaither oh ya that's all the same.

angaither:
I'm probably being super dense here, but where do you select the correct message with the correct contest id?

sbsmith86:
@angaithe no worries! 

[here](https://github.com/DoSomething/gladiator/pull/324/files?diff=split#diff-ee610ebae812c24f376dea1ff2170273R102) we set a `$params` array which we pass to as `sendEmail` function. 

[that function uses $params to grab the correct message](https://github.com/DoSomething/gladiator/blob/master/app/Services/Manager.php#L519)

angaither:
ah the ol' `where::($params)` of course 👍 


🚢 🚀 💌

app/Http/Controllers/MessagesController.php
```diff
@@ -92,7 +92,9 @@ public function sendMessage(Message $message)
     {
         // Get competition with activity.
         $competitionId = request('competition_id');
-        // $competition = null;
+
+        $contest = Contest::find(request('contest_id'));
+        $this->manager->appendCampaign($contest);
```
weerd: do you need to resave `$contest` here?

```php
$contest = $this->manager->appendCampaign($contest);
```

sbsmith86: not sure i know what you mean by resave? i need to append the campaign info...

weerd: it appends it sure, but it doesn't get actually added to the `$contest` unless you reassign `$contest` to the return value from `$this->manager->appendCampaign($contest);`

weerd: For example here: https://github.com/DoSomething/gladiator/blob/a924244398fe05108a1009867c04c9f3181f41f4/app/Http/Controllers/ContestsController.php#L86

weerd: Otherwise you're still passing along the original `$contest` data w/out the Campaign info attached.

sbsmith86: yea that is probably safer...but it actually works /shrug

I was following what was done here: 

https://github.com/DoSomething/gladiator/blob/f4fda9c508118e9051a60c532bf5d7041d7487a1/app/Http/Controllers/Api/UsersController.php#L96

I will update though.

sbsmith86: @weerd updated!

weerd: Shit. I think it's done incorrectly there too @_@ I don't know how that is working correctly...

sbsmith86: lemme just fix that while im here

weerd:
:sheep: it!

PR #332, opened 2016-08-02
Email Image toggle

### LINKED ISSUE #331: Toggle leaderboard images on/off
Allow admins to turn displaying images in the leaderboard email on/off easily.

sbsmith86:
#### What's this PR do?
Adds a checkbox to the message edit page that will toggle the display of images or not.

Go to a contest, edit the messages, toggle the "show images" check box on/off and then send yourself a test of a leaderboard message.

If it is "on" you should see reportback images, if it "off" you should not.

This is what the checkbox looks like:
image

This is how it looks in the DB: 
image

Fixes #331

### DIFF
--- a/app/Repositories/MessageRepository.php
+++ b/app/Repositories/MessageRepository.php
@@ -120,6 +120,10 @@ class MessageRepository
                 $data['type'] = $type;
                 $data['key'] = $key;
 
+                // @TODO - When the 'show_images' checkbox is not checked,
+                // The `show_images` property is not sent in the request so we
+                // manually set it here. Let's figure out a better way to do this,
+                // maybe in a MessageRequest class?
                 if ($data['type'] === 'leaderboard') {
                     if (! array_key_exists('show_images', $data)) {
                         $data['show_images'] = 0;
--- a/app/Repositories/MessageRepository.php
+++ b/app/Repositories/MessageRepository.php
@@ -121,7 +121,7 @@ class MessageRepository
                 $data['key'] = $key;
 
                 if ($data['type'] === 'leaderboard') {
-                    if (!array_key_exists('show_images', $data)) {
+                    if (! array_key_exists('show_images', $data)) {
                         $data['show_images'] = 0;
                     }
                 }
--- a/app/Repositories/MessageRepository.php
+++ b/app/Repositories/MessageRepository.php
@@ -120,6 +120,12 @@ class MessageRepository
                 $data['type'] = $type;
                 $data['key'] = $key;
 
+                if ($data['type'] === 'leaderboard') {
+                    if (!array_key_exists('show_images', $data)) {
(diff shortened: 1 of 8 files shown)


app/Repositories/MessageRepository.php
```diff
@@ -120,6 +120,12 @@ public function updateMessagesForContest($contest, $messages)
                 $data['type'] = $type;
                 $data['key'] = $key;
 
+                if ($data['type'] === 'leaderboard') {
+                    if (! array_key_exists('show_images', $data)) {
+                        $data['show_images'] = 0;
```
angaither: shouldn't this be the opposite? 
```
if ($show images)
// display the images
```
which means `show_images` should be 1, not zero? 


SORRY THAT THIS IS A STREAM OF BRAIN THOUGHTS AND MIGHT NOT MAKE SENSE.

sbsmith86: This is a weird hack i had to do because, for checkboxes, when you submit them unchecked the value doesn't get sent in the request. So what this does, is check if the `show_key` value was sent in the request, if so, then great use that value (1). If it wasn't sent, then that means the checkbox wasn't checked and we add the `show_images` key   to the request and set it to false.

sbsmith86: The other option is to do something like this, but i didn't like the hidden field thing http://nielson.io/2014/02/handling-checkbox-input-in-laravel/

angaither: ah, oy. who knew. ✅

angaither:
@sbsmith86 what does the email look like without sweet pics?

app/Repositories/MessageRepository.php
weerd: @sbsmith86 you can use a mutator! set it to something if not set! we do this already for other fields in Gladiator forms. Gimme a sec to find an example 😄

weerd: https://laravel.com/docs/5.2/eloquent-mutators#accessors-and-mutators from the docs and checking now for in Gladiator.

weerd: https://github.com/DoSomething/gladiator/blob/master/app/Models/Message.php#L53

sbsmith86:
@angaither 
image

sbsmith86:
Mentioned this in slack to @weerd but I don't think we can use mutators for this because the actual property is not sent in the request, we aren't trying to mutate as much as set it when it doesn't exist in the request...

See this for more details http://stackoverflow.com/questions/11424037/does-input-type-checkbox-only-post-data-if-its-checked

weerd:
👍 

Just add a `@TODO` for the checkbox weirdness for later investigation 🔍

sbsmith86:
@weerd TODO added! thanks!

PR #337, opened 2016-09-01
Added a 'share_link' token to emails.

### LINKED ISSUE #334: Pulling in Custom Social Share Link into Gladiator
@ngjo commented on [Thu Aug 11 2016](https://github.com/DoSomething/phoenix/issues/6871)

_FEATURE OVERVIEW_
-------------------
#### User Story
As an impact member, I would like the custom social share link to be made available as a mail merge field in Gladiator

#### Why This Matters
In talking to Mike about competitions, he (smartly) pointed out that it'd be useful to have this custom social share link included as a mail merge option in messaging. So, when I message a competitor, I'd like to reference their custom social share link in the emails so that it's easy for them to know what they should be doing. 

We would use this for any campaign that has competitions and custom social share links.

This could be:

- "Use your [custom social share link] and the more people you can get registered, the more chances at the $ you have!"

- Get more people to register via your link, here's a tweet for you: "Tweet this: "Register to vote at [custom social share link]" 

---

@angaither commented on [Fri Aug 12 2016](https://github.com/DoSomething/phoenix/issues/6871#issuecomment-239452265)

@ngjo this issue would actually belong in the gladiator repo then. 
Also since we are using the phoenix id for the social share, not the northstar id, gladiator doesn't have that variable. 

hmmm maybe we should be using NS id for the share link? 

---

@ngjo commented on [Fri Aug 12 2016](https://github.com/DoSomething/phoenix/issues/6871#issuecomment-239453122)

@angaither whoops, sorry about that! i wasn't sure how to get it in both voter-registration and gladiator...can it live in two places?

i think if it takes the northstar ID to pull into gladiator, that might be the best bet since we'll likely use these custom social share links in competition emails... what would it take to make the switch?

thanks!

angaither:
#### What's this PR do?
- this prefixes the links with /us/ so that when you hit phoenix a redirect doesn't prevent that query string from persisting.
- the token is `:share_link:`

Add `:share_link:` into an email and test the send, it should be populated with
`ds.org/us/node/[nid]?source=user/[id]`

- this is so gladiator emails can have those unique share links
- since we only have US competitions now (DoSomething/phoenix/issues/6849) I prefixed the link with `/us/` so this will NOT work if we run this on global things.

Fixes #334 

cc @ngjo

### DIFF
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -84,7 +84,8 @@ class Email
             ':reportback_noun:'       => strtolower($this->contest->campaign->reportback_info->noun),
             ':reportback_verb:'       => strtolower($this->contest->campaign->reportback_info->verb),
             ':sender_name:'           => $this->contest->sender_name,
-            ':rules_url:'            => ! is_null($this->competition) ? $this->competition->rules_url : '',
+            ':rules_url:'             => ! is_null($this->competition) ? $this->competition->rules_url : '',
+            ':share_link:'            => url(config('services.phoenix.uri') .'/us/node/' . $this->contest->campaign_id . '?source=user/' . $user->id),
         ];
 
         return $tokens;


sbsmith86:
👍

Issue #343, opened 2016-09-12
Showing contest level stats (summary of competition level stats)


ngjo:
#### User Story

As an admin, I have to report stats at both the contest and competition level. Right now, Gladiator shows stats at the competition level, but not the contest level:

https://gladiator.dosomething.org/competitions/83
![screen shot 2016-09-12 at 1 53 45 pm](https://cloud.githubusercontent.com/assets/2658867/18446564/e277bf26-78f0-11e6-8ea2-f368614774c6.png)

The Super Stress Face Contest (https://gladiator.dosomething.org/contests/17) has 3 separate competitions:

![screen shot 2016-09-12 at 1 59 03 pm](https://cloud.githubusercontent.com/assets/2658867/18446605/1cb839fe-78f1-11e6-8cad-0c8ef2b67102.png)

I would like to see the same Statistics to show up at the contest level and be the combination of the separate contests. Here are the statistics:
- Total number of contestants in competition:
- Number of contestants who have reported back:
- Reportback rate: 
- Approved reportbacks impact quantity: 
#### Why This Matters

Impact team reports these stats to campaign team and allows them to make appropriate tweaks to the competition wording. 
#### Current Workaround w/o Feature (if applicable)

Impact team manually copies and tracks the individual competition stats in a google doc and sums it up for the contest.

ngjo:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;bBfX8eaZ&#x2F;32-as-an-impact-team-member-i-want-to-see-contest-level-stats-so-that-i-can-quickly-see-how-an-entire-contest-is-doing-instead-of-p"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> As an impact team member, I want to see contest level stats so that I can quickly see how an entire contest is doing instead of pulling individual competition data.</a>

sbsmith86:
Admins are using looker for this so going to close for now.

PR #349, opened 2016-10-03
Contest stats

LindaOrtega:
#### What's this PR do?

Added contest-level statistics by aggregating competition statistics of specified contests. Tried to optimize performance by adding flash storage at contest level. 

Keep track of each competition statistic for a contest. Add up these competitions statistics. Verify that these manual calculations match the statistics displayed in the contest site. 

None.

Ref #343

### DIFF
--- a/app/Http/Controllers/ContestsController.php
+++ b/app/Http/Controllers/ContestsController.php
@@ -85,9 +85,26 @@ class ContestsController extends Controller
     {
         $contest = $this->manager->collectContestInfo($contest->id);
         $contest = $this->manager->appendCampaign($contest);
+        $competition_statistics = [];
+
+        foreach($contest->competitions as $competition) {
+          $key = generate_model_flash_session_key($competition, ['includeActivity' => true]);
+
+          if (session()->has($key)) {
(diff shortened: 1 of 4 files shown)


deadlybutter:
Managed to get my [old tool](https://github.com/DoSomething/usergenerator) working that generates a bunch of users + gladiator imports 

With 325 people that RB'd back for the entire content it took ~10 seconds to load. My guess here is [looping over all of these users](https://github.com/DoSomething/gladiator/blob/3b4d0527fbe472a60e76ea9aeacb1a04456eaad4/app/Services/Manager.php#L260) is whats taking so long. I think using flash storage is a good step to improve the performance but it seems like it might make more sense on the competition level where all of the expensive aggregation is taking place. 

If we really wanted to 💥 the complexity of this, computing all of this asynchronously & saving the results is the best bet. Have a button like "Calculate stats!" on the contest page which triggers a job to fire in the queue. The contest page would display whatever stats are currently saved & would require a refresh after the job finishes, but the app won't lockup while its computing. I hope that makes sense!

@weerd any thoughts / ideas you want to add?

PR #350, opened 2016-11-04
Leaderboard photos

### LINKED ISSUE #333: Option to edit the top 3 images in the leaderboard
#### User Story

As an admin, I'd like the option to change which photos within a user's profile makes it on to the top 3 sections of the leaderboard. I would still want the leaderboard to pull the most recent photo, but just with the option to change it (via the FID) if it's not the most representative photo.
#### Why This Matters

Often, the top 3 competitors are super engaged and post TONS of photos. Currently Gladiator auto pulls the most recent photo which sometimes is not a good representation of the complete impact that top 3 person has created. Then we get angry emails from the rest of the competition because the photo shown in the leaderboard sucks. Hopefully this wouldn't be an issue anymore if we're able to enter an alternate reportback item ID as the leaderboard photo(s).
#### Current Process

 Going into a competitor's profile and re-uploading a better photo they previously uploaded.

LindaOrtega:
#### What's this PR do?
For each of the top three users, this PR allows admin to choose which of the user's reportback item images will appear in leaderboard emails.

Navigate to a contest. Go to Email Settings. Check "Show Images" for all leaderboard emails that will be testing. Return to contest page and select a competition. Select "Set Photos" for a leaderboard email and enter the fid to feature for specific user. Should check the following test-scenarios:

     1. If Admin leaves fid field empty for user(s) and submits form, the db rows for those/that user(s) should not change and the test email should feature default (most recent) reportback item image for those/that user(s). 

     2. If Admin changes the fid for a user already present in db, the db row associated with that user should change. 

     3. If Admin enters incorrect fid for specific user/reportback, test email should show default reportback item image (the most recent one).

     4. If Admin enters correct fid for specified user/reportback, test email should show the selected reportback item image. 

     5. If Admin enters valid fid for specified user/reportback and for email type 1 but not for email type 2, the test email for email type 2 should not show the reportback item image entered for email type 1. 

PR composed of two parts: 
1) The process of editing/updating leaderboard_photos db
2) The process of composing leaderboard emails using the reportback item images present in this db. 

Fixes #333


app/Services/Manager.php
```diff
@@ -322,36 +323,92 @@ public function appendReportback($data, $parameters)
...
         $topThree = [];
+        $includeUserIds = false;
+        $hasCompetitionId = false;
+        $hasMessageId = false;
 
-        foreach ($topThreeUsers as $key => $user) {
-            $reportbackItems = $user->reportback->reportback_items->data;
-            $latestReportbackItem = array_pop($reportbackItems);
+        if (isset($options['includeUserIds']) && $options['includeUserIds']) {
+            $includeUserIds = true;
```
sbsmith86: I don't think you need to set a different variable here if it was already passed into the `$options` array instead, you could just replace the if statement on line 361 to with the if clause you have here.

Line 361 would then looks like:

```php
            // Provide info on user & reportback ids
             if (isset($options['includeUserIds']) && $options['includeUserIds']) {
                 $topThree[$key]['user_id'] = $user->id;
                 $topThree[$key]['reportback_id'] = $user->reportback->id;
             }
```

app/Services/Manager.php
```diff
@@ -322,36 +323,92 @@ public function appendReportback($data, $parameters)
...
+        }
+
+        if (isset($options['competition_id']) && isset($options['message_id'])) {
```
sbsmith86: Same here.

app/Services/Manager.php
```diff
@@ -322,36 +323,92 @@ public function appendReportback($data, $parameters)
...
+                $topThree[$key]['user_id'] = $user->id;
+                $topThree[$key]['reportback_id'] = $user->reportback->id;
+            }
+
+            // Provide image url/captaion of top three leaderboard images
+            if ($hasCompetitionId && $hasMessageId) {
+                $competition_id = $options['competition_id'];
+                $message_id = $options['message_id'];
+
+                $leaderboardReportbackItem = $this->getLeaderboardPhoto($competition_id, $message_id, $user->id);   //@NOTE calling Phoenix again
```
sbsmith86: Here to, you don't need to store a separate `$competition_id` variable you can just pass in the `$options` array values.

app/Services/Manager.php
```diff
@@ -322,36 +323,92 @@ public function appendReportback($data, $parameters)
...
     }
 
+    /*
+     * Determines whether a valid reportback_item _id is present in the LeaderboardPhotos DB
+     * for given competition, message, and leaderboard user.
+     * Returns the associated reportback item if present.
+     *
+     * @return  array $reportback_item/null
+     */
+    public function getLeaderboardPhoto($competition_id, $message_id, $user_id)
```
sbsmith86: These params should be 🐫  case (`$competitionId`). it's a laravel convention. variables/params are  🐫 case, database column are 🐍 case

app/Http/Controllers/CompetitionsController.php
```diff
@@ -248,4 +250,72 @@ public function updateFeaturedReportback(FeaturedReportbackRequest $request, Com
...
+     * Update the specified resource in storage.
+     *
+     * @param  \Gladiator\Requests\LeaderboardPhotosRequest  $request
+     * @param  \Gladiator\Models\Competition  $competition
+     * @param  \Gladiator\Models\Message  $message
+     * @return \Illuminate\Http\Response
+     */
+    public function updateLeaderboardPhotos(LeaderboardPhotosRequest $request, Competition $competition, Message $message)
+    {
+        foreach (range(0, 2) as $i) {
```
sbsmith86: interesting. Could you just use a regular `for` statement here, othewise it reads as a while loop, which is not what is happening. So:

`for ($i = 0; $i <= 2; $i++) {}` and then check that the `$reportback_item_id` is not null before setting the photo.

Want to shoot for readability here in case another developer jumps on this app.

app/Models/LeaderboardPhotos.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Gladiator\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class LeaderboardPhotos extends Model
```
sbsmith86: AH sorry i didn't catch this before. Convention is for models to be defined in the singular so this class should be `LeaderboardPhoto` instead

sbsmith86:
Looks great!! 👍 💯  go ahead and merge

Issue #181, opened 2016-04-07
Use shared Northstar-PHP library


DFurnes:
I could've sworn there was an issue for this already, but it'd be great to update Gladiator to use the shared [northstar-php](https://github.com/DoSomething/northstar-php) library. That way any issues we discover and fix here can also save us time when we're working on other services! :smile:

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [As a developer, I want Gladiator to be set up with Gateway](https://trello.com/c/neTDeFtF/193-as-a-developer-i-want-gladiator-to-be-set-up-with-gateway)

PR #351, opened 2016-12-02
Gateway Authentication of Users

LindaOrtega:
#### What's this PR do?
Allow admins to authenticate via Northstar using Gateway.

1. Clicking 'Log In' redirects user to Northstar authentication portal. 
2. If Admin, user should be able to sign in to Gladiator and see all contests. 
3. If Staff, user should also be able to sign in to Gladiator and see all contests. 
4. If unauthorized user tries to log in, all error handled in Northstar authentication portal. 

Gateway RestApiClient not implemented yet, so user activities currently crash. 

Ref #181


app/Http/Controllers/Auth/AuthController.php
```diff
@@ -18,30 +18,44 @@ class AuthController extends Controller
...
-    use AuthenticatesAndRegistersUsers, ThrottlesLogins;
-
+    
     /**
      * Where to redirect users after login / registration.
      *
      * @var string
      */
-    protected $redirectTo = '/';
+    //protected $redirectTo = '/';
```
sbsmith86: remove the commented out stuff :)

app/Http/Middleware/CheckRole.php
```diff
@@ -3,27 +3,38 @@
...
-        $roles = array_slice(func_get_args(), 2);
-
-        if (Auth::guest() || ! Auth::user()->hasRole($roles)) {
+        if ($this->auth->guest() || ! $this->auth->user()->hasRole($roles)) {
             if ($request->ajax()) {
                 return response('Unauthorized.', 401);
             } else {
-                // @TODO: respond with a more custom unauthorized access page or whatevs.
-                return redirect('/');
+                return view('/');
```
sbsmith86: I don't think you can use the `view()` function with a route, you would use it to serve a named view. So if you wanted to show the home page. you would do `view(pages.home)` 

But I think it might be fine to just leave the `return redirect('/')` instead.

config/services.php
```diff
@@ -14,10 +14,24 @@
     |
     */
 
+    // 'northstar' => [
```
sbsmith86: can remove.

database/seeds/UserTableSeeder.php
```diff
@@ -34,18 +34,11 @@ public function run()
     {
         // Add Admin Users
         $admins = [
```
sbsmith86: I actually think you should be able to remove all this work adding admin users now. so line 36-41 isn't needed since admins will be automatically created when they authenticate.

sbsmith86:
Looks like you have some style fixes to make, but looks good to me! When this gets merged and deployed to QA, in addition to regular testing, lets also take a look at the QA db and make sure it looks correct.

PR #356, opened 2017-01-23
Only remove a users from the waiting room being split

### LINKED ISSUE #326: Mysterious missing contestants from contest waiting room
The Contest [waiting room](https://gladiator.dosomething.org/contests/7) for the **Hot Dogs** contest seems to have had 118 (as of July 1, 2016) and then mysteriously dropped down to 93. All the missing users are valid contestants in the Gladiator DB with Northstar user account. Something seems to have happened where we dropped ~25 users from that particular waiting room.

This is something we should keep an eye on to see if it happens again and consider options for investigating what happened.

### LINKED ISSUE #346: Recreate competition not making it into waiting room bug
_TASK_
- Recreate the bug in this issue: https://github.com/DoSomething/devops/issues/177

### LINKED ISSUE #353: Delete user from correct waiting room
_BUG_
-------------------
#### Current Behavior

We have been noticing two different situations:

1. Users seeming to not make it into gladiator from phoenix. 
There were users in phoenix that had `competition=1` on their signup record. This means they clicked "Join" on the competition prompt when they signed up for a campaign. There was a discrepancy between the number of people who joined a competition on phoenix and the number of people in competitions in gladiator

2. It was flagged [over the summer](https://github.com/DoSomething/gladiator/issues/326) that users were going missing from waiting rooms. 

#### The Bug
The bug seems related to the splitting waiting room functionality. When an admin splits a waiting room we add them to a competition and then delete them from the waiting room.

[The bug is that we are deleting the users from all waiting rooms.](https://github.com/DoSomething/gladiator/blob/master/app/Models/WaitingRoom.php#L117)

The correct behavior is for the user to just be removed from the waiting room that got split. 

#### Steps to Replicate
Put a user into two different open waiting rooms and then split one, you will notice that the user gets removed from both waiting rooms instead of just the one that got split. 

#### Why This Matters

Because of this bug, people were not getting into competitions and therefore not receiving competitions messaging. 

[This was happening at a rate of 5-15%](https://dosomething.slack.com/archives/competitions/p1484860640000302)

#### Proposed Fix

Update the logic of deleting a user from a waiting rooms to only delete the user from the waiting room that just got split. This should clear up most of what we are seeing and if there are other issues sill occurring, we can pinpoint those separately. 

#### Other 

* We still should implement better monitoring in gladiator by getting it [set up with Gateway](https://github.com/DoSomething/gladiator/issues/181) so that we can track transaction ids of requests coming in and going out of the application.

* We should still implement stathat monitoring of spikes in the system to alert the team when stuff is looking wonky.

* We should still get better unit test coverage in the app.

sbsmith86:
#### What's this PR do?

Specifies which waiting room to remove a user from when splitting a waiting room into competitions.

Add a user to multiple waiting rooms and then split one of them. The user should only be removed from that waiting room being split not the other ones. 

[Here is all the context I think you need](https://github.com/DoSomething/gladiator/issues/353)

I'm sure there is some sort of allegory we can use to talk about this one little line of code and the havoc it caused, but I don't have one at the moment. Maybe it just represents a lesson in itself.

Fixes #353 

Also ...
Fixes #346 
Fixes #326

### DIFF
--- a/app/Models/WaitingRoom.php
+++ b/app/Models/WaitingRoom.php
@@ -97,6 +97,7 @@ class WaitingRoom extends Model
     {
         $startDate = Carbon::now()->startOfDay();
         $endDate = new Carbon($request->competition_end_date);
+        $waitingRoom = $contest->waitingRoom;
 
         foreach ($split as $competitionGroup) {
             // For each split, create a competition.
@@ -113,8 +114,8 @@ class WaitingRoom extends Model
             foreach ($competitionGroup as $userId) {
                 $user = User::find($userId);
 
-                // Remove them from the waiting room pivot table
-                $user->waitingRooms()->detach();
+                // Remove the user from the waiting room being split.
+                $user->waitingRooms()->detach($waitingRoom->id);
 
                 // Add them to the competitions pivot table
                 $user->competitions()->attach($competition->id);


weerd:
> I'm sure there is some sort of allegory we can use to talk about this one little line of code and the havoc it caused, but I don't have one at the moment. Maybe it just represents a lesson in itself.

We should turn it into an Aesop's Fable of sorts ;)

PR #360, opened 2017-02-07
Rename `id` column to `northstar_id`

sbsmith86:
#### What's this PR do?

Renames the `id` column on the `users` table to `northstar_id`. This will set us up transition to using Gateway as it expects a `northstar_id` to be on the users table.

Smoke test throughout the app, everything should work normally. Also run `phpunit` and tests should pass. 

Addresses #181

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -8,6 +8,13 @@ use Gladiator\Services\Northstar\Exceptions\NorthstarUserNotFoundException;
 class User extends BaseUser
 {
     /**
+     * The primary key for the model
+     *
+     * @var string
+     */
+    protected $primaryKey = 'northstar_id';
+
+    /**
      * Indicates if the IDs are auto-incrementing.
      *
      * @var bool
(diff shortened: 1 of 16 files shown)


app/Models/User.php
```diff
@@ -8,6 +8,13 @@
 class User extends BaseUser
 {
     /**
+     * The primary key for the model
+     *
+     * @var string
+     */
+    protected $primaryKey = 'northstar_id';
```
weerd: 🕺

database/seeds/UserTableSeeder.php
```diff
@@ -57,7 +57,7 @@ public function run()
 
             // Using first or create if someone is already an admin.
             $user = User::firstOrCreate([
-                 'id' => $contestant->id,
+                 'id' => $contestant->northstar_id,
```
weerd: Should this be:

```php
'northstar_id' => $contestant->id,
```

Since $contestant is a record from Northstar and it's creating a record into Gladiator?

weerd: Or do I have it backwards?

PR #370, opened 2017-02-17
User authentication in northstar using gateway

sbsmith86:
#### What's this PR do?

Authenticates users in northstar using gateway. 

This only does the work of using gateway for user authentication. Work for using gateway for other calls in coming in the next PR. 

We will no longer need to set admins for the app manually. If a user has an `admin` or `staff` role in northstar they should be able to get into gladiator. 

Addresses #181

### DIFF
--- a/resources/views/pages/home.blade.php
+++ b/resources/views/pages/home.blade.php
@@ -3,31 +3,17 @@
 @section('main_content')
 
     @include('layouts.header', [
-        'title' => 'All your base are belong to us!'
+        'title' => 'Gladiator'
     ])
 
     <div class="container -padded">
         <div class="wrapper">
             <div class="container__block -narrow">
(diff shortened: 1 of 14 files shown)


resources/views/pages/home.blade.php
```diff
@@ -3,31 +3,17 @@
...
         <div class="wrapper">
             <div class="container__block -narrow">
-                <p>In A.D. 2101</p>
-                <p>War was beginning.</p>
-                <p>Captain: What happen?</p>
-                <p>Mechanic: Somebody set up us the bomb.</p>
-                <p>Operator: We get signal.</p>
-                <p> Captain: What !</p>
-                <p>Operator: Main screen turn on.</p>
-                <p>Captain: It's you !!</p>
```
DFurnes: ‼️

PR #376, opened 2017-02-27
Wercker docker

blisteringherb:
#### What's this PR do?
I've updated the wercker.yml to use the ds-docker-php box as a build box to support the migration into the new world of Wercker deploys. I've also reverted the Capistrano deploys to work with Wercker.

Check Wercker to make sure the build and deploy works. https://app.wercker.com/dosomething/gladiator/runs


wercker.yml
```diff
@@ -1,34 +1,18 @@
-box: blisteringherb/scholarship@0.0.8
+box: blisteringherb/ds-wercker-box:mysql
```
DFurnes: This doesn't block merge, but it seems like the "Docker way" of handling service dependencies is attaching a [separate container](http://devcenter.wercker.com/docs/services) rather than having it all bundled in one. Maybe something to look into in a future PR?

pnasrat:
Great and based upon slack the goal is for this to move to the organization docker hub account at somepoint?

https://github.com/DoSomething/ds-docker-php

Confirmed passing run in wercker - and the changes look great to me

👍

wercker.yml
pnasrat: That seems like a good best practice to evaluate, but yeah shouldn't block this.

PR #377, opened 2017-03-03
Unsubscribe competition users

### LINKED ISSUE #358: Allow users to unsubscribe from competitions messaging.
Add a link to the bottom of all messaging that will allow a user to remove themselves from competitions messaging. 

When a user unsubscribes we only remove them from getting messaging for the competition that they unsubscribed for. The user is not removed from the competition. 

#### Tasks
* Add a column to the `competition_user` table to track if a user has unsubscribed from messaging.
* Filter users who have unsubscribed from the group of users we pull to send messages to
* generate a link that has the users id, and competition id, to unsubscribe the user from.

ScottPaillant:
#### What's this PR do?
This PR sets up the initial stages to allow members to unsubscribe from competitions messaging. Adds a new flag to allow for filtering competitions by those who are subscribed to them to get emails.

Assign unsubscribed true values to any user in local gladiator, and dd() $competition->subscribers when testing a message.

Initial steps to fix #358

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -136,11 +136,7 @@ class Manager
      */
     public function getModelUsers($model, $withReportback = false, $subscribed = false)
     {
-        if (! $subscribed) {
-            $users = $model->users;
-        } else {
-            $users = $model->subscribers;
-        }
+        $users = ! $subscribed ? $model->users : $model->subscribers;
 
         if (! $users) {
             return null;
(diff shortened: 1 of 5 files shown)


app/Services/Manager.php
```diff
@@ -134,9 +134,13 @@ public function collectContestInfo($data)
      * @param  bool  $withReportback
      * @return \Illuminate\Support\Collection  $users
      */
-    public function getModelUsers($model, $withReportback = false)
+    public function getModelUsers($model, $withReportback = false, $subscribed = false)
     {
-        $users = $model->users;
+        if (! $subscribed) {
```
sbsmith86: Think it would be cool to use a [ternary](https://davidwalsh.name/php-shorthand-if-else-ternary-operators) here:

`$users = ! $subscribed ? $model->users : $model->subscribers;`

PR #383, opened 2017-03-21
Creates unsubscribe api endpoint and functionality to expect a compet…

### LINKED ISSUE #379: Create and /unsubscribe endpoint
Create an endpoint in gladiator at `/api/v1/unsubscribe` that will take an northstar id and competition id and flag them as unsubscribed for that competition messaging.

Things needed: 

* Create a route in `Routes.php` in the `/api/v1` group at `/unsubscribe`
* Create an `UnsubscribeController` that has a `unsubscribe` method that gets hit when that new endpoint is hit.
* Create an `UnsubscribeRequest` class that gets passed to the `unsubscribe` method and validates the incoming request 
* In the `unsubscribe(UnsubscribeRequest $request)` method, gets the user, and sets them as unsubscribes for the competition that is passed in the request.
* Return a json response from `unsubscribe()` that confirms that unsubscribe was successful, or errors that something went wrong: ```{ message: 'success' } ```

ScottPaillant:
#### What's this PR do?
Creates unsubscribe api endpoint and functionality to expect a competition_id and northstar_id and unsubscribe user from competition_user pivot table

POST to gladiator.dev/api/v1/unsubscribe with a northstar_id, and a competition_id similiar to :
gladiator.dev/api/v1/unsubscribe?northstar_id=5894e9437f43c217dd158b49&competition_id=6
and check local dev to see if that user was unsubscribed (1 value in unsubscribed column in competition_user pivot table)

Fixes #379

### DIFF
--- a/app/Http/Controllers/Api/UnsubscribeController.php
+++ b/app/Http/Controllers/Api/UnsubscribeController.php
@@ -2,49 +2,35 @@
 
 namespace Gladiator\Http\Controllers\Api;
 
-use Gladiator\Services\Manager;
 use Gladiator\Models\Competition;
 use Gladiator\Http\Requests\UnsubscribeRequest;
 
 class UnsubscribeController extends ApiController
 {
     /**
-     * @var Gladiator\Services\Manager;
+     * Create new UnsubscribeController instance.
      */
-    protected $manager;
+    public function __construct()
+    {
+        $this->middleware('auth.api');
+    }
 
+    /**
+     * Unsubscribe user from competition.
+     *
+     * @param  Gladiator\Http\Requests\UnsubscribeRequest $request
+     * @return \Illuminate\Http\JsonResponse
+     */
     public function unsubscribe(UnsubscribeRequest $request)
     {
-        $credentials = $request->all();
+        $competition = Competition::find($request->input('competition_id'));
 
-        $competition = Competition::find($credentials['competition_id']);
+        $unsubscribed = $competition->unsubscribe($request->input('northstar_id'));
 
-        // Get competition with activity from flash if it is there.
-        // Otherwise, grab it.
-        $key = generate_model_flash_session_key($competition, ['includeActivity' => true]);
-
-        if (session()->has($key)) {
-            $competition = session($key);
-        } else {
-            $competition = $this->manager->getCompetitionOverview($competition, true, true);
-        }
-
-        $unsubscribed = $competition->unsubscribeUser($credentials['competition_id'], $credentials['northstar_id']);
-
-        if ($unsubscribed == 1) {
+        if ($unsubscribed) {
             return response()->json(['message' =>'success'], 200, [], JSON_UNESCAPED_SLASHES);
         } else {
             return response()->json(['message' =>'error'], 500, [], JSON_UNESCAPED_SLASHES);
         }
-    }
-
-    /**
-     * Create new UnsubscribeController instance.
-     */
-    public function __construct(Manager $manager)
-    {
-        $this->manager = $manager;
-
-        $this->middleware('auth.api');
     }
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/UnsubscribeController.php
```diff
@@ -0,0 +1,50 @@
...
+class UnsubscribeController extends ApiController
+{
+    /**
+     * @var Gladiator\Services\Manager;
+     */
+    protected $manager;
+
+    public function unsubscribe(UnsubscribeRequest $request)
+    {
+        $credentials = $request->all();
```
sbsmith86: I think it would be more readable to name this like `$data` or something. "credentials" sound like you are expecting a username/password to be sent in the request.

app/Http/Controllers/Api/UnsubscribeController.php
```diff
@@ -0,0 +1,50 @@
...
+
+        $competition = Competition::find($credentials['competition_id']);
```
sbsmith86: I don't think you need to load the whole competition here (or below) as you won't be doing anything with it.

app/Http/Controllers/Api/UnsubscribeController.php
ScottPaillant: For sure, I might've grabbed this when i was researching

app/Http/Controllers/Api/UnsubscribeController.php
ScottPaillant: We could sync on this, I was trying to find the best way to do an sql call, and I got the flow that it should be kept to the model, so I wasn't sure how to grab a Competition model to then do something like '$competition->unsubscribeUser'

app/Models/Competition.php
```diff
@@ -33,6 +33,14 @@ public function subscribers()
         return $this->belongsToMany(User::class, 'competition_user', 'competition_id', 'northstar_id')->wherePivot('unsubscribed', '=', 0);
     }
 
+    public function unsubscribeUser($competition_id, $northstar_id)
```
sbsmith86: You don't need to define the relationship again with `belongsToMany` since it has already been defined above. Also once you are in the `Competition` model you are already aware of the `id`and the `user`  (since they are properties/relationships on that instance of the model) so you shouldn't need to pass it here.

All you need to do is grab the record from the pivot table and update that 
This method could look more like this:
```
public function unsubscribe() {
  return DB::table('competitions_users')->where([
      ['northstar_id', '=', $this->user ],
      ['competition', '=', $this->id],
  ])->update(['unsubscribed' => 1]);
}
```

app/Http/Controllers/Api/UnsubscribeController.php
sbsmith86: Scratch that, you will need the model instance to call the `unsubscribe` method on it, but you don't need the activity to be attached to it. 

for example: 

```
$competition = Competition::find($data['competition_id']);
$competition->unsubscribe($data['northstar_id']);
```

app/Http/Controllers/Api/UnsubscribeController.php
```diff
@@ -0,0 +1,50 @@
...
+            return response()->json(['message' =>'success'], 200, [], JSON_UNESCAPED_SLASHES);
+        } else {
+            return response()->json(['message' =>'error'], 500, [], JSON_UNESCAPED_SLASHES);
+        }
+    }
+
+    /**
+     * Create new UnsubscribeController instance.
+     */
+    public function __construct(Manager $manager)
```
sbsmith86: You don't need the manager anymore so this constructor can be removed

sbsmith86:
Looks great!

Issue #381, opened 2017-03-16
BUG: Error on User Page


ScottPaillant:
On qa-gladiator, when viewing all contestants and clicking on a particular user an error is thrown:
![screen shot 2017-03-16 at 4 09 17 pm](https://cloud.githubusercontent.com/assets/9884758/24016731/ac8ed26e-0a63-11e7-8e07-368baaaa1a61.png)

Link: https://qa-gladiator.dosomething.org/users/55808092a59dbfa93d8b4568

Issue #380, opened 2017-03-16
BUG: Error on User page


sbsmith86:
On qa-gladiator, when clicking on a users name on a competition page, I get redirected to the users specific page and an error is thrown: 

![image](https://cloud.githubusercontent.com/assets/1700409/24012108/33fba1d4-0a53-11e7-977a-fe33f496db20.png)

sbsmith86:
NOTE: 

This profile works: https://qa-gladiator.dosomething.org/users/5571f4f5a59dbf3c7a8b4569

This profile does not: https://qa-gladiator.dosomething.org/users/559442cca59dbfc9578b4bf4

PR #384, opened 2017-03-28
Fixes messages bug where reportback attribute did not exist

ScottPaillant:
#### What's this PR do?
Change conditional value to check if what function will later use exists 

Local or QA-gladiator, go to a competition and click test welcome message 

what else?

#### What are the relevant tickets??
Fixes similar issue to #380 & #381 , pictures attached

screen shot 2017-03-28 at 2 43 57 pm

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -527,7 +527,7 @@ class Manager
     {
         $activity = $this->getActivityForUser($user->id, $parameters);
 
-        if ($activity) {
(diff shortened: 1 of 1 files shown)


sbsmith86:
> Local or QA-gladiator, go to a competition and click test welcome message

Can you try testing this with a send to a bulk list of users also...like the leaderboard message. (press 'send', not 'test' so that it actually tries to send to the group)

PR #389, opened 2017-04-03
Phoenix for activity

### LINKED ISSUE #388: BUG: Gladiator not pulling in user activity
_BUG_
-------------------
#### Current Behavior

When looking at a competition the stats/leaderboard show no activity for any campaign (even when users do have activity in phoenix)

#### Desired Behavior

We should be pulling in user activity from phoenix to be able to show the competition stats and leaderboard overview

#### Steps to Replicate
 
Go to any competition page and you will show it is reporting no reportback activity.

sbsmith86:
#### What's this PR do?

Switches calls to the northstar `/signups` proxy to calls directly to the phoenix API to grab user activity. 

You can review this PR commit by commit. The `AuthorizesWithDrupal` Trait and the added `Phoenix` service methods are borrowed (copied 😬 ) from Phoenix-next

We will have to do more thorough testing on QA and prod which have more users with activity. 

Fixes #388

### DIFF
--- a/app/Services/Phoenix/Phoenix.php
+++ b/app/Services/Phoenix/Phoenix.php
@@ -6,10 +6,15 @@ use DoSomething\Gateway\Common\RestApiClient;
 
 class Phoenix extends RestApiClient
 {
+    use AuthorizesWithDrupal;
+
+    /*
+     * Phoenix API base uri.
+     */
     protected $base_uri;
 
     /**
-     * Northstar constructor.
+     * Phoenix constructor.
      */
     public function __construct()
     {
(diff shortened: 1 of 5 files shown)


app/Services/Phoenix/Phoenix.php
```diff
@@ -55,4 +60,55 @@ public function getReportback($reportback_id, $reportback_item_id)
...
+     *
+     * @param array|string $query - query string, for filtering results
+     * @return array - JSON response
+     */
+    public function getAllReportbacks(array $query = [])
+    {
+        $path = 'reportbacks';
+        $query['load_user'] = true;
+        if (auth()->id()) {
+            $query['as_user'] = auth()->id();
```
weerd: Just to be sure, this will query as the currently signed in user and not specifying it as a specific member?

sbsmith86: @weerd I'm not sure what this is for. honestly. It looks like yes, it will be the currently signed in user which should be fine since only admins are allowed to log in. But also we aren't really using this method.

weerd: @sbsmith86 Gotcha. Ok, cause then it'll just be getting RBs that the admin user submitted. Just something to note for future reference 💃

PR #394, opened 2017-04-10
Leaderboard bug

### LINKED ISSUE #392: BUG: Leaderboard showing n/a for all constestants
### Current Behavior

Competitions are showing `n/a` for all values and contestants:

Seems to be happening on only some competitions, For example: 

http://gladiator.dosomething.org/competitions/245/leaderboard

screen shot 2017-04-10 at 9 39 36 am

### Expected behavior

This should pull in activity for the users.

sbsmith86:
#### What's this PR do?

Now that we are getting user activity directly from phoenix, the data gets returned to us with embedded arrays instead of objects. This changes how we pull in the data to the leaderboard view and the leaderboard export. 

This PR updates the code so that we pull in the data accessing the arrays, instead of trying to use objects. 

* Go to a leaderboard and you should see the data in the view instead of `N/A`.
* Try to export a leaderboard and it should export properly without throwing errors that you are trying to get a property of a non-object. 

Fixes #392

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -77,23 +77,23 @@ class Manager
             ];
 
             if (isset($user->reportback)) {
-                $details[] = reportback_admin_url($user->reportback->id);
-                $details[] = $user->reportback->quantity;
-                $details[] = $user->reportback->reportback_items->count_by_status['promoted'];
-                $details[] = $user->reportback->reportback_items->count_by_status['approved'];
-                $details[] = $user->reportback->reportback_items->count_by_status['excluded'];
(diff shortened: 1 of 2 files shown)


app/Services/Manager.php
```diff
@@ -116,7 +116,7 @@ public function getStatus($data, $status)
         $list = [];
 
         foreach ($data as $reportbackitem) {
-            if ($reportbackitem->status == $status) {
+            if ($reportbackitem['status'] == $status) {
```
weerd: Can we update this to `===`? :)

sbsmith86: word.

PR #400, opened 2017-04-26
One per competion

### LINKED ISSUE #396: Only Allow Users to be In 1 competition per contest.
_FEATURE OVERVIEW_
-------------------
#### User Story

Only allow a user to be in one competition per contest. Add a check in `[UsersController@store`](https://github.com/DoSomething/gladiator/blob/master/app/Http/Controllers/Api/UsersController.php#L68) that checks if a user is in a competition for that contest first before adding the user to a waiting room. 

If the user is already signed up, return an error response: 
```
{
  "error": {
    "code": 422,
    "message": "User Already In Competition"
  }
}
```

#### Additional Information (optional)

We have an api endpoint at `POST /users` that adds users to a waiting room for a competition. This is used by other apps to add people to competitions. 

Currently, a user can only being in a waiting room once. However, if that waiting room gets split into competitions, then they can be added back to the waiting room and therefore can be split into another competition. So it is possible that a user could get into multiple competitions for one contest. 

#### Why This Matters

In phoenix-next, if a user gets prompted to join a competition twice then we only want them to be able to actually join once.

ScottPaillant:
#### What's this PR do?
This PR creates the functionality to check if a user is in a competition for a contest before adding them to the waiting room for the contest (and subsequently a 2nd competition)

Attempt to add a user that already exists in a competition for a contest to a waiting room for that contest

Fixes #396

### DIFF
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -619,13 +619,11 @@ class Manager
     public function findUserInContest($contest, $id)
     {
         $userCompetitions = $contest->competitions->filter(function ($competition, $key) use ($id) {
-
             $users = $competition->users->where('northstar_id', $id);
 
             if (! $users->isEmpty()) {
                 return $competition;
             }
-
         });
 
         return $userCompetitions->isEmpty() ? null : $userCompetitions;
--- a/app/Services/Manager.php
+++ b/app/Services/Manager.php
@@ -618,14 +618,15 @@ class Manager
      */
     public function findUserInContest($contest, $id)
     {
-        $userCompetitions = $contest->competitions->filter(function ($competition, $key) use ($id) {
-            $users = $competition->users->where('northstar_id', $id);
+        // Collection of competitions the user belongs to in contest
+        $result = $contest->competitions->filter(function ($competition, $key) use ($id) {
+            $user = $competition->users->where('northstar_id', $id);
 
-            if (! $users->isEmpty()) {
+            if (! $user->isEmpty()) {
                 return $competition;
             }
         });
 
-        return $userCompetitions->isEmpty() ? null : $userCompetitions;
+        return $result->isEmpty() ? null : $result;
     }
 }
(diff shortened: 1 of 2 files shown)


weerd:
Is it possible that a user could be in a waiting room (not split into a contest yet) and be placed again into the waiting room if requested again?

sbsmith86:
@weerd nope, i double checked that.

app/Services/Manager.php
```diff
@@ -608,4 +608,24 @@ public function sendEmail($user, $contest, $params)
...
+     * Finds if user exists in a competition for a given contest
+     *
+     * @param app\Models\Contest         $contest
+     * @param string                     $id
+     * @return array|null
+     */
+    public function findUserInContest($contest, $id)
+    {
+        $userCompetitions = $contest->competitions->filter(function ($competition, $key) use ($id) {
+            $users = $competition->users->where('northstar_id', $id);
```
weerd: Maybe rename this from `$users` to `$result`? It's a bit confusing because it seems like it'll return more than one user, but this is find a specific single user in an series of potential competitions they might be in, it just happens to return a collection or whatever.

app/Services/Manager.php
```diff
@@ -608,4 +608,24 @@ public function sendEmail($user, $contest, $params)
...
+
+            if (! $users->isEmpty()) {
+                return $competition;
+            }
+        });
+
+        return $userCompetitions->isEmpty() ? null : $userCompetitions;
```
weerd: Kinda same thing w/ `$userCompetitions`. Should only be one, but happens to be in collection. Might be easier to read if more generic name? thoughts?

app/Services/Manager.php
ScottPaillant: I Agree! It should only ever be one ($users and $userCompetitions), but it (userCompetitions) does have the capability to hold more. I can make it more clear

sbsmith86: @weerd ya we "technically" only allow one person per competition, but we haven't had any actual logic enforcing that until now, so I thought it was better to not make the assumption that we would only get one thing back.

PR #401, opened 2017-04-27
Updating /get user endpoint to return either a competition or a waitingRoom

### LINKED ISSUE #399: Update the API for Phoenix-Next requests
Phoenix-Next (PN) will be adding a component on the campaign page to allow users to signup for a competition once they affiliate with the campaign. For the time being this component will be a persistent item at the top of the Action page regardless of whether the user is signed up or not.

We need Gladiator to expose a couple things for PN via the API, such as asking if a user is signed up for a contest (making sure to account for a user being in a waiting room OR already moved into a competition). PN will be making a get request to the users endpoint and pass along the following parameters: 
```
/api/v1/users?user=58flongnorthstarid99&campaign_id=1144&campaign_run_id=2234
```
I believe that's the minimum amount of info Gladiator would need to figure out if the specified user is in a contest, by filtering via the `users index` endpoint.

Also, we need to make sure that Gladiator checks to confirm a user is not already in a contest (waiting room or competition; but maybe checking against the contest is enough) since PN will for the time being only use local storage in the browser to know if a user signed up for a competition or not. It is _very_ possible that if the local storage is cleared or the user is on a new browser, then they could click to signup for a competition again. Previously, Phoenix-Ashes has a DB table that knew whether a user was signed up or not, so no extra requests would get sent to Gladiator. Since PN for the time being isn't doing that, Gladiator should be double-checking (probs should be doing this anyhoo!).

Here is the [work already done on PN](https://github.com/DoSomething/phoenix-next/pull/222) to enable making API requests to Gladiator if needed for reference!

ScottPaillant:
#### What's this PR do?

Created `/get user` api endpoint that can be passed an `id`, `campaign_id`, and a `campaign_run_id` and returns a user with either the waiting room they are in or the competition

Example response when user is in waiting room:
``` {
  "data": {
    "id": "5894e9437f43c217dd158b49",
    "first_name": null,
    "last_name": null,
    "email": null,
    "mobile": null,
    "signup": null,
    "reportback": null,
    "created_at": "2017-03-16T18:34:50+00:00",
    "updated_at": "2017-04-27T14:30:41+00:00",
    "waitingRoom": {
      "data": {
        "id": 33,
        "competition_start_date": {
          "date": "2016-04-25 00:00:00.000000",
          "timezone_type": 3,
          "timezone": "UTC"
        },
        "competition_end_date": {
          "date": "2016-05-30 00:00:00.000000",
          "timezone_type": 3,
          "timezone": "UTC"
        }
      }
    }
  }
} 
```

Example response when user is in a competition:
```
  "data": {
    "id": "5722361ba59dbfb8528b4595",
    "first_name": null,
    "last_name": null,
    "email": null,
    "mobile": null,
    "signup": null,
    "reportback": null,
    "created_at": "2016-04-28T16:11:06+00:00",
    "updated_at": "2016-04-28T16:11:06+00:00",
    "competition": {
      "data": {
        "id": 13,
        "competition_start_date": {
          "date": "2016-04-28 00:00:00.000000",
          "timezone_type": 3,
          "timezone": "UTC"
        },
        "competition_end_date": {
          "date": "2016-05-30 23:59:59.000000",
          "timezone_type": 3,
          "timezone": "UTC"
        }
      }
    }
  }
```
Issue a GET request to api/v1/users with the parameters `id`, `campaign_id`, and `campaign_run_id` and you should receive a user in JSON format with appropriate data.

See issue #399 

Fixes #399

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -69,7 +69,7 @@ class UsersController extends ApiController
                 $user->setAttribute('roomAssignment', $roomAssignment);
             }
 
-            $competitionAssignment = $this->manager->findUserInCompetition($contest , $user->northstar_id);
+            $competitionAssignment = $this->manager->findUserInCompetition($contest, $user->northstar_id);
 
             // If a user is in a competition for a given contest attach the competition to the users object
             if ($competitionAssignment) {
@@ -77,7 +77,6 @@ class UsersController extends ApiController
             }
 
             return $this->item($user);
-
         } else {
             return response()->json([
                 'error' => [
@@ -153,7 +152,8 @@ class UsersController extends ApiController
      * @param int                   $campaign_run_id
      * @return app\Models\Contest
      */
-    public function getContest($campaign_id, $campaign_run_id) {
+    public function getContest($campaign_id, $campaign_run_id)
+    {
         $contest = Contest::with(['waitingRoom', 'competitions'])->where('campaign_id', '=', $campaign_id)
                             ->where('campaign_run_id', '=', $campaign_run_id)
                             ->firstOrFail();
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -53,10 +53,39 @@ class UsersController extends ApiController
      *
      * @return \Illuminate\Http\Response
      */
-    public function index()
+    public function index(UserRequest $request)
     {
-        // @TODO: temp for now, likely want to use repository and transform.
-        return User::all();
+        $data = $request->all();
+
+        $user = User::find($data['id']);
+
+        $contest = $this->getContest($request['campaign_id'], $request['campaign_run_id']);
+
+        if ($contest) {
+            $roomAssignment = $user->waitingRooms()->find($contest->waitingRoom->id);
+
+            // If a user is in a waiting room attach the waitingRoom to the users object
+            if ($roomAssignment) {
+                $user->setAttribute('roomAssignment', $roomAssignment);
+            }
+
+            $competitionAssignment = $this->manager->findUserInCompetition($contest , $user->northstar_id);
+
+            // If a user is in a competition for a given contest attach the competition to the users object
+            if ($competitionAssignment) {
+                $user->setAttribute('competitionAssignment', $competitionAssignment);
+            }
+
+            return $this->item($user);
+
+        } else {
+            return response()->json([
+                'error' => [
+                    'code' => 404,
+                    'message' => 'Contest Not Found',
+                ],
+            ]);
+        }
     }
 
     /**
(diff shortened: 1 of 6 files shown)


app/Http/Transformers/UserTransformer.php
```diff
@@ -8,13 +8,24 @@
...
+
+    /**
      * Transform resource data.
      *
      * @param  User  $user
      * @return array
      */
     public function transform(User $user)
     {
+        // dump("UserTransformer here");
```
sbsmith86: remove comment

app/Services/Manager.php
```diff
@@ -612,21 +613,16 @@ public function sendEmail($user, $contest, $params)
     /**
      * Finds if user exists in a competition for a given contest
      *
-     * @param app\Models\Contest         $contest
-     * @param string                     $id
-     * @return array|null
+     * @param app\Models\Contest                $contest
```
sbsmith86: just need one space or tab

app/Http/Controllers/Api/UsersController.php
```diff
@@ -53,10 +53,38 @@ public function __construct(Registrar $registrar, Manager $manager, UserReposito
...
      * @return \Illuminate\Http\Response
      */
-    public function index()
+    public function index(UserRequest $request)
     {
-        // @TODO: temp for now, likely want to use repository and transform.
-        return User::all();
+        $data = $request->all();
+
+        $user = User::find($data['id']);
```
weerd: Should this be in a conditional if no `id` is passed?

app/Http/Controllers/Api/UsersController.php
```diff
@@ -53,10 +53,38 @@ public function __construct(Registrar $registrar, Manager $manager, UserReposito
      *
      * @return \Illuminate\Http\Response
      */
-    public function index()
+    public function index(UserRequest $request)
     {
-        // @TODO: temp for now, likely want to use repository and transform.
-        return User::all();
+        $data = $request->all();
```
weerd: Not sure if you really need to separate this into a `$data` variable... it's only really used below for finding the user, but then you switch to `$request['param_name']`. Would recommend just sticking w/ `$request['...']` approach throughout or `$request->input('...')`, etc :)

app/Http/Controllers/Api/UsersController.php
ScottPaillant: Id is required in UserRequest!

app/Http/Controllers/Api/UsersController.php
ScottPaillant: Makes sense!

sbsmith86:
@weerd we can talk tomorrow morning if the response format makes sense for you, or what you were thinking. You will either get back a `waitingRoom` or `Competition` property on the response and that will indicate that the user has already been assigned something. I am going to approve this for now. If you need updates tomorrow morning. I am on it!

app/Http/Controllers/Api/UsersController.php
```diff
@@ -53,10 +53,36 @@ public function __construct(Registrar $registrar, Manager $manager, UserReposito
      *
      * @return \Illuminate\Http\Response
      */
-    public function index()
+    public function index(UserRequest $request)
     {
-        // @TODO: temp for now, likely want to use repository and transform.
-        return User::all();
+        $user = User::find($request['id']);
```
weerd: This isn't vital for merge but something to consider. In REST an `index` endpoint typically always returns a collection/array of items, in this case users, and allows for filtering it down _if_ parameters are passed.

sbsmith86: @weerd we can tackle that tomorrow when @ScottPaillant gets in but I think this unblocks you?

weerd:
Sure thing, sounds good :) We can chat further tomorrow if needed. I left an additional comment but it's not a merge blocker by any means. More of a recommendation and not necessarily something needed for now.

app/Http/Controllers/Api/UsersController.php
weerd: [yep yep!](https://github.com/DoSomething/gladiator/pull/401#issuecomment-297850746)

PR #406, opened 2017-05-04
Add env variable, and new unsubscribe link token

ScottPaillant:
#### What's this PR do?
Added token for Email's being sent in gladiator to include unsubscribe_link crafted with users id and competition id.

Send a test message and check if :unsubscribe_link: is properly replaced.

Ongoing progress to create unsubscribe link in gladiator

### DIFF
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -87,6 +87,7 @@ class Email
             ':sender_name:'           => $this->contest->sender_name,
             ':rules_url:'             => ! is_null($this->competition) ? $this->competition->rules_url : '',
             ':share_link:'            => url(config('services.phoenix.uri') .'/us/node/' . $this->contest->campaign_id . '?source=user/' . $user->id),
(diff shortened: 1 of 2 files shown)


app/Services/Email.php
```diff
@@ -87,6 +87,7 @@ protected function defineTokens($user)
             ':sender_name:'           => $this->contest->sender_name,
             ':rules_url:'             => ! is_null($this->competition) ? $this->competition->rules_url : '',
             ':share_link:'            => url(config('services.phoenix.uri') .'/us/node/' . $this->contest->campaign_id . '?source=user/' . $user->id),
+            ':unsubscribe_link:'      => url(config('services.northstar.profile_url') . '/unsubscribe?' . 'northstar_id='.$user->id . '&competition_id=' . $this->competition->id),
```
sbsmith86: can you do something like ` url(config('services.northstar.profile_url', ['northstar_id' => $user->id, 'competition_id' => $competition->id])`? I think that will build out the url for you automatically.

ScottPaillant: For sure!

PR #409, opened 2017-05-09
Display user info in leaderboard messages

sbsmith86:
#### What's this PR do?

In #408, we fixed and issue sending leaderboard messages. However there still seemed to be an issue with the content of the message not pulling in the user names to the leaderboard. Example: 

image

I am not fully sure when this was introduced, but it seems it might have been introduced when we started using Gateway for making calls to northstar. These calls return `NorthstarUser` objects, which we then appended activity to and cataloged for the leaderboard. Essentialy the `leaderboard` variable we sent to the message view was full of `NorthstarUser`objects. 

We are using Mail Queues to send these message, and queues do not allow objects to be passed to the view, it converts them to arrays, and it was stripping the `NorthstarUser` data when it did that. 

The work around I came up with was to just create a smaller array with only the data we needed to use to display in the leaderboard. It's not pretty and, it's a bit of a hack. Going to try to schedule some time into my next sprint to clean some of this stuff up. 

Leaderboard messages now: 

image

We need to deploy to QA and TESSST

Addresses #408

### DIFF
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -196,11 +196,27 @@ class Email
     {
         $leaderboard = $this->competition->activity['active'];
 
+        // @TEMP - Convert the NorthstarUser objects that make up the leaderboard array into smaller arrays with just the information we need to display in the email. This is a hacky way of getting around an issue where we can't pass serialized objects to the message view through the mail queue.
+        $entryArray = [];
+
+        foreach ($leaderboard as $user) {
+            $entry = [];
+
+            $entry = [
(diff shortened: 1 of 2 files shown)


weerd:
Just curious but would casting the `(array) NorthstarUser` object to an array help with resolving the references?

sbsmith86:
@weerd no that didn't seem to work I think because not all of the properties in the array are public.

weerd:
Gotcha, ok. Good to know!

PR #412, opened 2017-05-31
Unsubscribe link in emails

sbsmith86:
#### What's this PR do?
This PR actually adds the unsubscribe link to each of the email templates. 

Send yourself a test email and it should have a unsubscribe link in the bottom.

### DIFF
--- a/resources/views/messages/checkin.blade.php
+++ b/resources/views/messages/checkin.blade.php
@@ -1,3 +1,3 @@
 <p>{!! $content['body'] !!}</p>
 
-<p><a href="{!! $content['unsubscribe_link'] !!}">Unsubscribe</a></p>
+<p>Don’t want to receive emails about this competition? <a href="{{ $content['unsubscribe_link'] }}">Unsubscribe</a></p>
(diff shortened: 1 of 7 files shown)


resources/views/messages/checkin.blade.php
```diff
@@ -1 +1,3 @@
-{!! $content['body'] !!}
+<p>{!! $content['body'] !!}</p>
+
+<p><a href="{!! $content['unsubscribe_link'] !!}">Unsubscribe</a></p>
```
DFurnes: I'm pretty sure this can be safely escaped! `{{ $content['unsubscribe_link'] }}`

resources/views/messages/welcome.blade.php
```diff
@@ -1 +1,3 @@
-{!! $content['body'] !!}
+<p>{!! $content['body'] !!}</p>
```
DFurnes: Is this Markdown? If so, is it already wrapped in `<p></p>` tags?

sbsmith86: No it's not markdown, i need to `<p></p>` tags to make sure the body and the unsubscribe link show up on different lines.

sbsmith86:
@DFurnes updated! Can I get your 👀 again?

resources/views/messages/welcome.blade.php
DFurnes: Got it!

PR #413, opened 2017-06-13
Check if there are no subscribers

sbsmith86:
#### What's this PR do?

Checks if we get back an empty collection when we grab subscribers. 

On QA, find a competition or create one where there is only one person in the competition, then unsubscribe. Send a message and nothing should be sent. 

Without this check, it would default send to everyone in the competition, which is the exact opposite of what we want.


weerd:
> Without this check, it would default send to everyone in the competition, which is the exact opposite of what we want.

😬

PR #418, opened 2017-08-04
Fix for leaderboard photos not updating

### LINKED ISSUE #416: Bug: Updating Leaderboard photos doesn't work
When editing the top three leaderboard photos, we are not saving the values into the DB. 

It looks like we are still trying to look for and update a `user_id` column and everything has been updated to use `northstar_id` 

https://github.com/DoSomething/gladiator/blob/master/app/Http/Controllers/CompetitionsController.php#L290-L318

sbsmith86:
#### What's this PR do?

Uses `northstar_id` instead of `user_id` to find the Leaderboard photo to update and to store in the DB. 

👀 

Fixes #416

### DIFF
--- a/app/Http/Controllers/CompetitionsController.php
+++ b/app/Http/Controllers/CompetitionsController.php
@@ -290,21 +290,21 @@ class CompetitionsController extends Controller
     public function updateLeaderboardPhotos(LeaderboardPhotoRequest $request, Competition $competition, Message $message)
     {
         for ($i = 0; $i <= 2; $i++) {
-            // request format: _method, _token, user_id_{{$index}},
+            // request format: _method, _token, northstar_id_{{$index}},
             //                 reportback_id_{{$index}}, reportback_item_id_{{$index}}
-            $userId = $request->input('user_id_'.$i);
+            $northstarId = $request->input('northstar_id_'.$i);
             $reportbackId = $request->input('reportback_id_'.$i);
             $reportbackItemId = $request->input('reportback_item_id_'.$i);
 
             // If none of ids null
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/CompetitionsController.php
```diff
@@ -290,21 +290,21 @@ public function editLeaderboardPhotos(Competition $competition, Message $message
...
             //                 reportback_id_{{$index}}, reportback_item_id_{{$index}}
-            $userId = $request->input('user_id_'.$i);
+            $northstarId = $request->input('northstar_id_'.$i);
             $reportbackId = $request->input('reportback_id_'.$i);
             $reportbackItemId = $request->input('reportback_item_id_'.$i);
 
             // If none of ids null
-            if (($userId != 0) && ($reportbackId != 0) && ($reportbackItemId != 0)) {
-                $photo = LeaderboardPhoto::where('competition_id', '=', $competition->id)->where('message_id', '=', $message->id)->where('user_id', '=', $userId)->first();
+            if (($northstarId != 0) && ($reportbackId != 0) && ($reportbackItemId != 0)) {
```
DFurnes: This check can probably be removed, since Northstar doesn't have a `0` anonymous user like Drupal.

DFurnes: Oh unless this is a null check? If so, could we just check truthiness or explicitly against `is_null`?

sbsmith86: @DFurnes ah yea i think it is a null check because it is a form request we are sending over, so we are making sure it was sent.

DFurnes: Hmm yeah, okay! I'm fine with merging as-is.

PR #419, opened 2017-08-07
only show top 100 in email

sbsmith86:
#### What's this PR do?

Slices the leaderboard to the top 100 folks before rendering it in the email. 

Send a leaderboard and you should only see the top 100 in the email. 

I originally thought this would help with some performances bottlenecks that happen when trying to grab large amount of RB data for a large set of users. But that was stupid because we still need all the user activity to calculate who is in the top 100 

image

Regardless, it does seem really heavy to put the full leaderboard, that is sometimes over 1k people, in one email, so I think chopping it makes sense for the user experience anyway. 

Addresses https://www.pivotaltracker.com/story/show/149209601

### DIFF
--- a/app/Services/Email.php
+++ b/app/Services/Email.php
@@ -201,6 +201,8 @@ class Email
     {
         $leaderboard = $this->competition->activity['active'];
 
+        $leaderboard = array_slice($leaderboard, 0, 100);
+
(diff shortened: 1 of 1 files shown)


DFurnes:
2017, vintage! Closing this out so I don’t keep getting GitHub reminders to review it.

(Hi Shae!!! 👋)

PR #424, opened 2017-09-11
add content-length header to api routes

sbsmith86:
#### What's this PR do?

Adds the `content-length` header to api responses.

Make an api request and see the `content-length` header in the set of response headers.

Certain api requests were returning invalid JSON. 

For example, https://gladiator.dosomething.org/api/v2/contests?page=22 , returns invalid JSON.

I used https://jsonlint.com/ to validate the response object and it is invalid, returning this error:

```
Error: Parse error on line 154:
...d469c644e398bfa97", "588e90d2a0bfad499f5
-----------------------^
Expecting 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '[', got 'undefined'
``` 

It looks like it is trying to return the over 2000 users in one of the competitions under the contest and it cuts off the response, making the response invalid JSON. Digging into why this might happen and it looks like if we set the `content-length` header on the response it will specify how long the body of the response should be. 

Open to other suggestions!

Fixes https://www.pivotaltracker.com/story/show/150406511

### DIFF
--- a/app/Http/Middleware/AddContentLength.php
+++ b/app/Http/Middleware/AddContentLength.php
@@ -3,6 +3,7 @@
 namespace Gladiator\Http\Middleware;
 
 use Closure;
+use Illuminate\Support\Str;
 
 class AddContentLength
 {
@@ -16,7 +17,7 @@ class AddContentLength
     public function handle($request, Closure $next)
     {
         $response = $next($request);
-        $response->header('Content-Length', strlen(serialize($response)));
+        $response->header('Content-Length', Str::length($response->getContent()));
 
         return $response;
     }
--- a/app/Http/Middleware/AddContentLength.php
+++ b/app/Http/Middleware/AddContentLength.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Gladiator\Http\Middleware;
+
+use Closure;
+
+class AddContentLength
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/AddContentLength.php
```diff
@@ -0,0 +1,23 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $response = $next($request);
+        $response->header('Content-Length', strlen(serialize($response)));
```
DFurnes: I think we want `$response->getContent()`, since the `serialize()` function would give us the length of the PHP serialized object, not the HTTP response itself. And let's use either `mb_strlen` or Laravel's `Str::length` so that we don't run into issues with UTF-8 characters in the response (like emoji! ✨)

```php
$response->header('Content-Length', Str::length($response->getContent()));
```

sbsmith86:
@DFurnes I was thinking about reaching out to @blisteringherb about any server settings we could change. I can't figure out how to make the response shorter because quasar wants to ingest everything we have in gladiator and there are competitions with a lot of users in them. Short of just providing total counts, I'm not sure what we can do there.

sbsmith86:
@DFurnes updated to get the length of the actual HTTP response

DFurnes:
@sbsmith86 Oh hmm, I think I might have actually lied on the `Str::length` usage - the `Content-Length` header should be [giving the length in bytes](http://mark.koli.ch/remember-kids-an-http-content-length-is-the-number-of-bytes-not-the-number-of-characters), not "actual" characters, so `str_len` is probably the right usage here. Sorry! 😬

